PR #22, opened 2015-10-23
Reorganize key names

dcsjapan:
See [flarum/core #265](https://github.com/flarum/core/issues/265).

- Adjusts key names to three-tier namespacing.
- Extracts previously unextracted strings.
- Code fix needed in DiscussionTaggedPost.js before this can be used.


dcsjapan:
Incidentally, once you've fixed the code in **DiscussionTaggedPost.js** you can close https://github.com/flarum/core/issues/527. :)

PR #23, opened 2015-11-10
Fixing Issue flarum/core#330

dbohn:
This pull request is not finished yet, but it is, that you can follow our progress.

For details about the steps we want to take, please refer to the issue flarum/core#330

### DIFF
--- a/js/admin/src/addTagChangePermission.js
+++ b/js/admin/src/addTagChangePermission.js
@@ -1,0 +1,27 @@
+import { extend } from 'flarum/extend';
+import PermissionGrid from 'flarum/components/PermissionGrid';
+import SettingDropdown from 'flarum/components/SettingDropdown';
+
+export default function() {
+  extend(PermissionGrid.prototype, 'startItems', items => {
+    items.add('allowTagChange', {
+      icon: 'tags',
+      label: app.translator.trans('flarum-tags.admin.permissions.allow_edit_tags_label'),
+      setting: () => {
+        const minutes = parseInt(app.settings.allow_tag_change, 10);
+
+        return SettingDropdown.component({
+          defaultLabel: minutes
+            ? app.translator.transChoice('core.admin.permissions_controls.allow_some_minutes_button', minutes, {count: minutes})
+            : app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button'),
+          key: 'allow_tag_change',
+          options: [
+            {value: '-1', label: app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button')},
+            {value: '10', label: app.translator.trans('core.admin.permissions_controls.allow_ten_minutes_button')},
+            {value: 'reply', label: app.translator.trans('core.admin.permissions_controls.allow_until_reply_button')}
+          ]
+        });
+      }
+    }, 90);
+  });
+}
--- a/src/Access/DiscussionPolicy.php
+++ b/src/Access/DiscussionPolicy.php
@@ -10,10 +10,12 @@
 
 namespace Flarum\Tags\Access;
 
+use Carbon\Carbon;
 use Flarum\Core\Access\AbstractPolicy;
 use Flarum\Core\Discussion;
 use Flarum\Core\User;
 use Flarum\Event\ScopeHiddenDiscussionVisibility;
+use Flarum\Settings\SettingsRepositoryInterface;
 use Flarum\Tags\Tag;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Database\Eloquent\Builder;
@@ -21,10 +23,25 @@ use Illuminate\Database\Query\Expression;
 
 class DiscussionPolicy extends AbstractPolicy
 {
+
     /**
      * {@inheritdoc}
      */
     protected $model = Discussion::class;
+
+    /**
+     * @var SettingsRepositoryInterface
+     */
+    protected $settings;
+
+    /**
+     * @param SettingsRepositoryInterface $settings
+     * @param Gate $gate
+     */
+    public function __construct(SettingsRepositoryInterface $settings)
+    {
+        $this->settings = $settings;
+    }
 
     /**
      * {@inheritdoc}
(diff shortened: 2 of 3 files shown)


franzliedke:
Awesome, looking good already!

It looks like too much got removed in the js/admin/dist/extension.js file, though. I'm assuming this is because you did not run `bower install` in the js/admin folder. Which is not documented, I know. ;)

dbohn:
We think the issue is fixed now. If there are any problems, please let us know, so that we can fix them.

Additionally we have created a pull request for the flarum/english repository to add a translation field for the new setting.

js/admin/dist/extension.js
```diff
@@ -412,6 +412,40 @@ $.fn.sortable = function(options) {
 return sortable;
 }));
 ;
+System.register('flarum/tags/addTagChangePermission', ['flarum/extend', 'flarum/components/PermissionGrid', 'flarum/components/SettingDropdown'], function (_export) {
```
tobscure: Please remove this file from the PR, as we only commit the dist files when releasing a new version. See https://github.com/flarum/flarum/blob/master/CONTRIBUTING.md#pull-request-guidelines

js/admin/src/addTagChangePermission.js
```diff
@@ -0,0 +1,29 @@
...
+import PermissionGrid from 'flarum/components/PermissionGrid';
+import SettingDropdown from 'flarum/components/SettingDropdown';
+
+export default function() {
+  extend(PermissionGrid.prototype, 'startItems', items => {
+    items.add('allowTagChange', {
+      icon: 'tags',
+      //label: app.translator.trans('core.admin.permissions.allow_changing_tags_label'),
+      // TODO: Add Translation String here
+      label: 'Allow changing tags',
```
tobscure: Not using the English translation / it doesn't match what you've PR'd to flarum/english?

src/Access/DiscussionPolicy.php
```diff
@@ -10,23 +10,39 @@
...
+use Flarum\Settings\SettingsRepositoryInterface;
 use Flarum\Tags\Tag;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Query\Expression;
 
-class DiscussionPolicy extends AbstractPolicy
-{
+class DiscussionPolicy extends AbstractPolicy {
+
```
tobscure: Watch out for PSR-2 here

src/Access/DiscussionPolicy.php
```diff
@@ -52,10 +68,21 @@ public function before(User $actor, $ability, Discussion $discussion)
...
+            if ($ability == 'tag') {
+                $actorCanEditTags = $this->actorCanEditTags($actor, $discussion);
+            }
+
             foreach ($tags as $tag) {
+
+                if ($ability == 'tag' && ($actor->hasPermission('tag' . $tag->id . '.discussion.tag') || $actorCanEditTags)) {
+                    return true;
+                }
+
```
tobscure: Instead of putting logic in the `before` method, try adding a new method named according to the ability (i.e. `tag`). The AbstractPolicy class automatically calls the appropriately-named method if it exists. See flarum/core's PostPolicy@edit method as an example. Let me know if this doesn't make sense

tobscure:
Looking good, thanks!

I've added a few comments on the diff, let me know if you have any questions :)

dbohn:
thanks, missed to push a commit that addresses two of your issues. Will look at the other two. Thanks for pointing these points out!

dbohn:
Okay, the other two issues are also fixed. Thanks for pointing these out.

Please let us know, if there is anything missing or not working as expected. If everything is fine, I'll squash the commits.

src/Access/DiscussionPolicy.php
```diff
@@ -103,4 +120,25 @@ public function scopeHiddenDiscussionVisibility(ScopeHiddenDiscussionVisibility
...
+        if ($discussion->start_user_id == $actor->id) {
+            $allowEditTags = $this->settings->get('allow_tag_change');
+
+            return ($allowEditTags === '-1'
+                || ($allowEditTags === 'reply' && $discussion->participants_count <= 1)
+                || ($discussion->start_time->diffInMinutes(new Carbon) < $allowEditTags)
+            );
+        }
+
+        return false;
```
tobscure: I haven't tested but my feeling is that the `return false` is unnecessary. Abilities are denied by default and only granted if `true` is returned; not returning false will allow further logic could be added by another extension for example.

Other than that, perfect!

dbohn: I have looked into the Flarum\Core\Access\Gate class and I think that it is only possible for before-Policies to be chained.

Otherwise the "resolveAuthCallback" method is called which does execute at least one policy method. So it is not possible to chain multiple policies for the same permission.

I could remove it though. It is unnecessary as PHP treats null as false if not compared with a strict comparison. I have written it, because I had to write a lot of Java in the last time, where you have to add a return statement at the end of every code path in every non-void method, and just got used to it.

tobscure: Flarum's Policies are actually implemented in a different way to the Laravel way (which is where the Flarum\Core\Access\Gate class comes from):

1. A single `before` handler is registered on the `Gate` in `CoreServiceProvider`.
2. Within this handler, the `GetPermission` event is called.
3. `AbstractPolicy` (and thus every Policy class) listens for this event and calls a method named after the ability if it exists.
4. If a non-null value is returned, execution of further listeners stops.

So a lot of the code in the Gate class is not actually used (I should clean that up sometime!)

dbohn:
I squashed the commits now. Hope everything is still alright.
Thanks for your great support while we developed the fix. It was a great experience for us.

I'm eager to participate more in Flarum by developing extensions and stuff, but first I have to finish my bachelor thesis.

tobscure:
Merged! Thanks for contributing – we really appreciate it! Good luck with your thesis, and hope to see you around in the future :)

PR #28, opened 2016-04-14
Added REST tags endpoint

nielstholenaar:
See  https://github.com/flarum/core/issues/600 for more information.

### DIFF
--- a/src/TagRepository.php
+++ b/src/TagRepository.php
@@ -41,7 +41,7 @@ class TagRepository
      */
     public function all(User $user = null)
     {
-        $query = Tag::newQuery();
+        $query = Tag::query();
 
         return $this->scopeVisibleTo($query, $user)->get();
     }
--- a/src/Api/Controller/ListTagsController.php
+++ b/src/Api/Controller/ListTagsController.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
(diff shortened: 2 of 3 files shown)


src/TagRepository.php
```diff
@@ -17,6 +17,16 @@
 class TagRepository
 {
     /**
+     * Get a new query builder for the users table.
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function query()
+    {
+        return Tag::query();
```
franzliedke: Why is this `query` when we used `newQuery` below?

franzliedke:
This is looking very good. Just one little question.

src/TagRepository.php
nielstholenaar: Because the tag model doesn't has an newQuery method.

nielstholenaar:
I've chosen to use the `Tag::query()` method instead of the `Tag::newQuery()`.

Mainly because the `Tag::query()


The newQuery method errors out with the message: `Call to undefined method Flarum\Tags\TagRepository::newQueryWithoutScopes()`

tobscure:
Also ListTagController should be ListTag**s**Controller :)

nielstholenaar:
@tobscure I've renamed it :-)

tobscure:
Awesome! Oh, I just noticed that this will spit out all tags regardless of permissions. I think `$this->tags->all($request->getAttribute('actor'));` should work? If you can test it out a bit that'd be great :)

franzliedke:
I've done that now.

Thanks so much @nielstholenaar!

PR #29, opened 2016-08-01
Add posts API filter for tags

coderstephen:
Add the means of filtering post queries by the tags of the posts' parent discussions. See https://github.com/flarum/core/issues/990.


franzliedke:
:+1:

PR #30, opened 2016-08-28
Set title of the Tag Page

terabin:
All Tag Pages have the same forum Title, it's a good idea change the title when you visit a new tag.

### DIFF
--- a/js/forum/src/components/TagHero.js
+++ b/js/forum/src/components/TagHero.js
@@ -4,8 +4,6 @@ export default class TagHero extends Component {
   view() {
     const tag = this.props.tag;
     const color = tag.color();
-
-    app.setTitle(tag.name());
     
     return (
       <header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')}
@@ -18,5 +16,9 @@ export default class TagHero extends Component {
         </div>
       </header>
     );
+  }
+  config() {
+    const tag = this.props.tag;
+    app.setTitle(tag.name());
   }
 }


tobscure:
Thanks for the PR! We have an issue for this somewhere so that's good :)

Just one thing: the `app.setTitle` call should be made in the `config` method rather than `view`. This is because a call to `view` constructs the virtual dom, but doesn't guarantee that the page is actually being displayed - whereas `config` is only called once the page has definitely been rendered. See `IndexPage` in core as an example.

terabin:
Just like this? The parameter `IsInitialized` is required?

terabin:
Just a question, ClientView set the title for search engines, so, the correct way to do this is changing the php file too right?

Another point, if ClientView set first title, why the function setTitle is called when app start?

tobscure:
Yes, that's correct, setting the title in PHP is necessary to set the title for search engines. It's necessary to set the title in both PHP and JS because Flarum is a single-page application, meaning users can navigate between a tag page and a discussion, and then back to a tag page, without hitting the server for new HTML.

So this is a little more tricky... The route for a tag page is defined in `Listener\AddClientAssets` on line 55. Currently it does not point to a custom controller, which means it'll just use the default `WebAppController` and then when the JS app boots it will see the current URL and load the correct discussion list. A new controller called `TagController` needs to be created which extends `WebAppController`, injects a `TagsRepository` instance, and overrides the `getView` method:

```php
    protected function getView(Request $request)
    {
        $view = parent::getView($request);

        $slug = array_get($request->getQueryParams(), 'slug');
        $actor = $request->getAttribute('actor');
        $tag = $this->tags->query()->whereVisibleTo($actor)->where('slug', $slug)->firstOrFail();

        $view->setTitle($tag->name);

        return $view;
    }
```

Ultimately this controller should extend `IndexController` and preload the discussion list too, but let's not worry about that for now.

Also, something else I didn't notice on my first pass: in the JS you're setting the title in the `TagHero` component, but the presence of this component is not technically synonymous with being on a tag page... The title should really be set in the extension of the `IndexPage` component in `addTagFilter.js`:

```js
  extend(IndexPage.prototype, 'config', function(value, isInitialized) {
    if (isInitialized) return;

    const tag = this.currentTag();

    if (tag) {
      app.setTitle(tag.name());
    }
  });
```

Sorry to be nitpicky, but it's good to do these things right :)

terabin:
I'm studying the Flarum, then surely I will need advice, thanks!

But I did not understand well, my Core does not have the same structure as the current Core ,although the latest version .

Currently, information is passed to the app.blade by ClientView.php in render function.

The correct would be to override this function, create TagRepository and consult the data by the slug, is not it?

tobscure:
I'm guessing you're using beta 5 - but the ClientView code has been completely refactored in dev-master. You'll need to update if you want to want to contribute :)

franzliedke:
@terabin Do you need further help to get started with the PHP side of things?

terabin:
@franzliedke, any help surely be welcome.
I have studied and endeavored to understand the Flarum. I hope that soon may make great contributions.

luceos:
@flarum/core what to do with this? As suggested on discord#staff I'd like to propose to close stale PR's based on old code.

tobscure:
I'm closing this due to inactivity. It's almost workable from the JS side but the `app.setTitle` call should really be made as an extension of `IndexPage.prototype.config` rather than a `TagHero` (which should not be coupled to the document title).

From the PHP side of things, the explanation in my comment above still stands.

PR #33, opened 2016-12-02
Adjust namespacing

dcsjapan:
- Updates namespacing of all translation keys.
- Namespacing of settings is unchanged.

### DIFF
--- a/js/lib/helpers/tagLabel.js
+++ b/js/lib/helpers/tagLabel.js
@@ -25,7 +25,7 @@ export default function tagLabel(tag, attrs = {}) {
   return (
     m((link ? 'a' : 'span'), attrs,
       <span className="TagLabel-text">
-        {tag ? tag.name() : app.translator.trans('flarum-tags.lib.deleted_tag_text')}
+        {tag ? tag.name() : app.translator.trans('tags::deleted_tag_text')}
       </span>
     )
   );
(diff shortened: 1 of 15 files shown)


dcsjapan:
I think I've got it right this time. 😃

js/lib/helpers/tagLabel.js
```diff
@@ -25,7 +25,7 @@ export default function tagLabel(tag, attrs = {}) {
   return (
     m((link ? 'a' : 'span'), attrs,
       <span className="TagLabel-text">
-        {tag ? tag.name() : app.translator.trans('flarum-tags.lib.deleted_tag_text')}
+        {tag ? tag.name() : app.translator.trans('tags::deleted_tag_text')}
```
franzliedke: @dcsjapan So `lib` can be removed like `admin` as well?

Just asking for confirmation as this seems to be the only exception...

franzliedke:
Thank you very much! All changes look syntactically correct (just waiting for a reply to that one question). I did not check whether you missed anything, but if so, we'll notice that later anyway. :)

issyrocks12:
Yes they can be removed, same as admin and forum

dcsjapan:
Closed as @tobscure has concluded removal of the scope from key names isn't worth the extra fuss.

franzliedke:
@dcsjapan Did I miss that? Where did he say so?

tobscure:
@franzliedke in a private discussion with @dcsjapan

dcsjapan:
Sorry, I could have phrased that more clearly!

PR #36, opened 2017-04-06
Fix Internal Server Error for blank API requests

dshoreman:
Attempting to send an empty body to the create and update endpoints will result in Flarum outputting "Internal Server Error" as plain text, instead of failing gracefully with a json-ised validation exception.

This adds an empty array as the third parameter to `array_get()`, which sets the default value and thus allows Flarum to do its thing.

Response before this change:

Response after applying fix:


tobscure:
Thanks!

PR #37, opened 2017-04-06
Fix ordering bug

CDK2020:
If an empty request is sent to the /tags/order endpoint, all tags would get moved to secondary tags.

Sorry for lots of commits, I'm not used to StyleCI

### DIFF
--- a/src/Listener/UpdateTagMetadata.php
+++ b/src/Listener/UpdateTagMetadata.php
@@ -11,6 +11,7 @@
 
 namespace Flarum\Tags\Listener;
 
+use Flarum\Core\Post;
 use Flarum\Event\DiscussionWasDeleted;
 use Flarum\Event\DiscussionWasStarted;
 use Flarum\Event\PostWasDeleted;


src/Api/Controller/OrderTagsController.php
```diff
@@ -30,6 +30,10 @@ public function handle(ServerRequestInterface $request)
 
         $order = array_get($request->getParsedBody(), 'order');
 
+        if ($order == '') {
```
franzliedke: Can you change this to `empty($order)`, please, in case this is an empty array?

Also, we can supply a third argument to the `array_get()` call two lines above to have an empty array as default value.

tobscure:
@franzliedke but an empty array should be allowed to mean that there are no primary tags, all tags are secondary...?

src/Api/Controller/OrderTagsController.php
tobscure: @franzliedke but an empty array should be allowed to set all tags as secondary?

issyrocks12: I like that, empty array sets all to secondary, but still prevents accidents.

franzliedke: I don't get where the empty string comes from, though. The PR description talks about an "empty request". That would mean that `array_get($request->getParsedBody(), 'order')` returns `null` (its default value for missing keys). And `null == ''` (as we are not doing strict comparison). How about `is_null` then?

CDK2020: I'm still going to advocate for my way because is_null has function call overhead.

franzliedke: Then do `$order === null`, please.

franzliedke:
@CDK2020 Are you still interested in completing this? If so, please take care of the requested change. :smile:

luceos:
@franzliedke @CDK2020 has other responsibilities from RL at the moment. I guess we can safely assume he won't be able to conclude the PR.

PR #38, opened 2017-04-11
Specify available includes for /api/tags endpoint

dshoreman:
Currently the list endpoint doesn't return the parent relation, making it difficult to work with child tags. Attempts to add `?include=parent` to the URL return an empty `errors=[]` response.

With this PR, `/api/tags` will include a child tag's parent under `relationships`.
Appending `?include=parent,lastDiscussion` will also include the latest discussion.

### DIFF
--- a/src/Api/Controller/ListTagsController.php
+++ b/src/Api/Controller/ListTagsController.php
@@ -25,6 +25,20 @@ class ListTagsController extends AbstractCollectionController
     public $serializer = TagSerializer::class;
 
     /**
+     * {@inheritdoc}
+     */
+    public $include = [
+        'parent',
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public $optionalInclude = [
+        'lastDiscussion',
+    ];
+
+    /**
      * @var \Flarum\Tags\Tag
      */
     protected $tags;


tobscure:
Thanks!

franzliedke:
@tobscure Does `$include` only load the relationship information (i.e. adding the `relationship` subfield to the corresponding tag resource), or does it include the full resource in the `included` field of the JSON-API response?

dshoreman:
@franzliedke It only includes the ID as a relationship. Example output from Postman:

```json
{
  "data": [
    {
      "type": "tags",
      "id": "5",
      "attributes": {
        "name": "API Test",
        "description": "A tag to test the API with",
        "slug": "api-test",
        "color": null,
        "backgroundUrl": null,
        "backgroundMode": null,
        "iconUrl": null,
        "discussionsCount": 0,
        "position": 0,
        "defaultSort": null,
        "isChild": true,
        "isHidden": false,
        "lastTime": "2017-04-07T12:02:05+00:00",
        "canStartDiscussion": false,
        "canAddToDiscussion": false
      },
      "relationships": {
        "lastDiscussion": {
          "data": {
            "type": "discussions",
            "id": "14"
          }
        },
        "parent": {
          "data": {
            "type": "tags",
            "id": "1"
          }
        }
      }
    }
  ]
}
```

franzliedke:
Okay. That's a bit confusing, because of JSON-API's [`include` query parameter](http://jsonapi.org/format/#fetching-includes)...

dshoreman:
I'm not sure where the difference lies - `POST /tags` will include the full `included` section with the same `$include = ['parent'];` line added in the controller.

dshoreman:
@franzliedke Adding a `->load('parent')` seems to include the full data, although I'm currently running beta.7 so it'll be slightly different for master. Would you like a PR to include full `included` section?

tobscure:
@dshoreman that'd be great, thanks!

dshoreman:
@franzliedke @tobscure Sorry guys, derp moment. I must've scrolled straight past the `  "included": [` line earlier and thought the included parents were part of the main data object. Full output is included for both parent and last discussion.

Tested on beta.7 with my tweaks, as well as a clean master clone.

I should sleep... 😴

tobscure:
PR would still be welcome, as currently the included relations are not being eager loaded → n+1 queries. See some of the API List controllers in flarum/core for examples on how to do this (basically just `->load($this->getIncluded($request))`, from memory)

PR #39, opened 2017-04-11
Enable parent to be set during creation via API

dshoreman:
Sets the parent ID as long as it exists and is a top-level primary tag.
If it's not, it'll simply be ignored. Position is set automatically to be one higher than the current highest, or 0 if parent has no children.

Example data (in addition to the usual attributes):
```
"relationships": {
    "parent": {
        "data": {
            "id": 1
        }
    }
}
```

### DIFF
--- a/src/Api/Controller/CreateTagController.php
+++ b/src/Api/Controller/CreateTagController.php
@@ -26,6 +26,11 @@ class CreateTagController extends AbstractCreateController
     public $serializer = TagSerializer::class;
 
     /**
+     * {@inheritdoc}
+     */
+    public $include = ['parent'];
+
+    /**
      * @var Dispatcher
      */
     protected $bus;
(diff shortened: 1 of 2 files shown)


dshoreman:
@tobscure Not sure if you were referring to order or parent in the Gitter chat. Since I had to write this for my project, figured I'd submit in case it's useful to someone. You're welcome to close if you'd rather wait until the overhaul

tobscure:
This looks good. I'm not sure about is the fact that setting parent to `0` makes it top-level primary, whereas leaving it `null` makes it secondary — perhaps not the most obvious API behaviour. Also, for complete correctness, there should be an exception thrown if an invalid parent ID is provided.

However, given that this will all be reviewed when the overhaul is done, I'm happy to merge this now :) Thanks!

franzliedke:
Thanks for your contributions, @dshoreman ! :handshake:

PR #40, opened 2017-04-11
Lazy-load any included relationships on /api/tags

dshoreman:
Adds eager-loading to the tags list API as per discussion in #38


tobscure:
:beers:

PR #41, opened 2017-05-18
Show modal if # of primary/secondary tags is less than required

datitisev:
Shows a modal if the number of primary or secondary tags is less than the number specified in settings.
I've gotten the filter for the primary and secondary tags from `js/admin/src/TagsPage.js` ([L60](https://github.com/flarum/flarum-ext-tags/blob/master/js/admin/src/components/TagsPage.js#L60), [L69](https://github.com/flarum/flarum-ext-tags/blob/master/js/admin/src/components/TagsPage.js#L69))
Fixes flarum/core#1101. Please squash ^_^

### Changelog
- [Fixed] Errored if num of primary/secondary tags is less than required

### DIFF
--- a/js/forum/src/addTagComposer.js
+++ b/js/forum/src/addTagComposer.js
@@ -45,7 +45,7 @@ export default function() {
   override(DiscussionComposer.prototype, 'onsubmit', function(original) {
     if (!this.tags.length
       || (!this.tags.filter(tag => tag.position() !== null && !tag.isChild()).length < app.forum.attribute('minPrimaryTags'))
-      || (!this.tags.filter(tag => tag.position() === null || tag.isChild()).length < app.forum.attribute('minSecondaryTags'))) {
+      || (!this.tags.filter(tag => tag.position() === null).length < app.forum.attribute('minSecondaryTags'))) {
       app.modal.show(
         new TagDiscussionModal({
           selectedTags: this.tags,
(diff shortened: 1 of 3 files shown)


js/forum/src/addTagComposer.js
```diff
@@ -43,10 +43,12 @@ export default function() {
   });
 
   override(DiscussionComposer.prototype, 'onsubmit', function(original) {
-    if (!this.tags.length) {
+    if (!this.tags.length
+      || (!this.tags.filter(tag => tag.position() !== null && !tag.isChild()).length < app.forum.attribute('minPrimaryTags'))
+      || (!this.tags.filter(tag => tag.position() === null).length < app.forum.attribute('minSecondaryTags'))) {
```
franzliedke: @datitisev Are you sure this is working correctly? To me it looks like the negation in the beginning negates the result of `this.tags.filter(...).length`, which does not make sense. If you are planning to negate the `<` comparison, then I'd suggest just using the `>=` operator.

Also, please extract these conditions into variables with a sensible name, so that it is easier to understand what this code does.

datitisev: Oh, yeah... Whoops. Sorry about that.
I'll try to fix that whenever I have time, hopefully in about 10 hours.
Als, I think it did work in my testing, but looking at it now, all I can see it'd be is `false <` or `true <`... I will have to revisit this.
Thanks for the review :D

datitisev: @franzliedke Done! I wasn't sure if I should use `chosenPrimaryTags` or just `primaryTags`.

datitisev:
How is it now? Sorry, I committed the fixes a month ago, and forgot to comment.

franzliedke:
Thanks - good job! :handshake:

datitisev:
No problem! 😁

PR #42, opened 2017-06-28
$negate implementation in TagGambit

ar7n:
$negate implementation in TagGambit

### DIFF
--- a/src/Gambit/TagGambit.php
+++ b/src/Gambit/TagGambit.php
@@ -43,24 +43,39 @@ class TagGambit extends AbstractRegexGambit
     {
         $slugs = explode(',', trim($matches[1], '"'));
 
-        // TODO: implement $negate
-        $search->getQuery()->where(function ($query) use ($slugs) {
+        $search->getQuery()->where(function ($query) use ($slugs, $negate) {
             foreach ($slugs as $slug) {
                 if ($slug === 'untagged') {
-                    $query->orWhereNotExists(function ($query) {
-                        $query->select(new Expression(1))
-                              ->from('discussions_tags')
-                              ->where('discussions.id', new Expression('discussion_id'));
-                    });
+                    if (!$negate) {
+                        $query->orWhereNotExists(function ($query) {
+                            $query->select(new Expression(1))
+                                ->from('discussions_tags')
+                                ->where('discussions.id', new Expression('discussion_id'));
+                        });
+                    } else {
+                        $query->orWhereExists(function ($query) {
+                            $query->select(new Expression(1))
+                                ->from('discussions_tags')
+                                ->where('discussions.id', new Expression('discussion_id'));
+                        });
+                    }
                 } else {
                     $id = $this->tags->getIdForSlug($slug);
-
-                    $query->orWhereExists(function ($query) use ($id) {
-                        $query->select(new Expression(1))
-                              ->from('discussions_tags')
-                              ->where('discussions.id', new Expression('discussion_id'))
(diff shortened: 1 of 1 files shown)


src/Gambit/TagGambit.php
```diff
@@ -43,24 +43,39 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
...
+                                ->from('discussions_tags')
+                                ->where('discussions.id', new Expression('discussion_id'));
+                        });
+                    } else {
+                        $query->orWhereExists(function ($query) {
+                            $query->select(new Expression(1))
+                                ->from('discussions_tags')
+                                ->where('discussions.id', new Expression('discussion_id'));
+                        });
+                    }
```
franzliedke: You can simplify this by passing `$negate` as the [`orWhereExists` method's second parameter](https://github.com/laravel/framework/blob/5823a994a0ad05ba90bcfa02e5747113fd26ca2e/src/Illuminate/Database/Query/Builder.php#L1195). I would typically recommend against that, as boolean parameters are less clear, but here the variable is clearly named and it shortens our code, so that should be fine.

src/Gambit/TagGambit.php
```diff
@@ -43,24 +43,39 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
...
+                                ->where('tag_id', $id);
+                        });
+                    } else {
+                        $query->orWhereNotExists(function ($query) use ($id) {
+                            $query->select(new Expression(1))
+                                ->from('discussions_tags')
+                                ->where('discussions.id', new Expression('discussion_id'))
+                                ->where('tag_id', $id);
+                        });
+                    }
```
franzliedke: Same here, of course.

ar7n:
@franzliedke Done

src/Gambit/TagGambit.php
```diff
@@ -43,24 +43,23 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
...
             foreach ($slugs as $slug) {
                 if ($slug === 'untagged') {
                     $query->orWhereNotExists(function ($query) {
                         $query->select(new Expression(1))
-                              ->from('discussions_tags')
-                              ->where('discussions.id', new Expression('discussion_id'));
-                    });
+                            ->from('discussions_tags')
+                            ->where('discussions.id', new Expression('discussion_id'));
+                    }, $negate);
```
franzliedke: This is not correct. `orWhereNotExists()` does not accept that second `$negate` parameter, only `orWhereExists()` does. Please change to use this method, and thus change to `!$negate` as well.

src/Gambit/TagGambit.php
```diff
@@ -43,24 +43,23 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
...
 
                     $query->orWhereExists(function ($query) use ($id) {
                         $query->select(new Expression(1))
-                              ->from('discussions_tags')
-                              ->where('discussions.id', new Expression('discussion_id'))
-                              ->where('tag_id', $id);
-                    });
+                            ->from('discussions_tags')
+                            ->where('discussions.id', new Expression('discussion_id'))
+                            ->where('tag_id', $id);
```
franzliedke: Also, please don't change the indentation.

franzliedke:
@ar7n Are you still interested in fixing this pull request?

franzliedke:
@ar7n I've rebased and cleaned up your commit and pushed it onto master.

Thanks for the PR! :heart:

PR #46, opened 2018-02-01
Fix issue #1249 and do not show or count private discussions in Tags …

michaeldalyuk:
This is to fix: flarum/core#1249


tobscure:
Thanks @michaeldalyuk, this looks good to me. Will leave for @franzliedke or @luceos to merge :)

luceos:
I'm curious why we are persisting the count to the database. Would we not do that we could return a value based on the permissions, is that something we should consider for the future @flarum/core ?

tobscure:
@luceos performance

PR #48, opened 2018-04-01
Added default <meta> description for tags

johannsa:
Continuing the work on flarum/core#1385, this is adding the <meta> description for tags. If a tag does not have a description, it would default to _"Discussions with the '{tag}' tag"_.

### DIFF
--- a/js/forum/src/components/TagHero.js
+++ b/js/forum/src/components/TagHero.js
@@ -7,7 +7,10 @@ export default class TagHero extends Component {
     app.setTitle(tag.name());
     let description = tag.description();
     if (!description) {
-      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() }).join('');
+      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() });
+      if (description.constructor === Array) {
+        description = description.join('');
+      }
     }
     app.setDescription(description);
 
--- a/js/forum/src/components/TagHero.js
+++ b/js/forum/src/components/TagHero.js
@@ -4,6 +4,12 @@ export default class TagHero extends Component {
   view() {
     const tag = this.props.tag;
     const color = tag.color();
+    app.setTitle(tag.name());
+    let description = tag.description();
+    if (!description) {
+      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() }).join('');
+    }
+    app.setDescription(description);
 
     return (
       <header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')}
--- a/src/Forum/Controller/TagController.php
+++ b/src/Forum/Controller/TagController.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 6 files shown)


franzliedke:
Don't commit dist files, please.

johannsa:
The compiled files were already deleted on a second commit I made minutes after the original one, @franzliedke.

franzliedke:
I still see them marked as being changed in [the diff](https://github.com/flarum/flarum-ext-tags/pull/48/files).

They should not be deleted, just changes should not be committed.

johannsa:
I see, I committed them by mistake at the beginning and then I made a second commit deleting them.

Any advise on how to proceed from here or should I just revert to initial status, unstage the files and then force the push?

franzliedke:
Revert the removal, squash your commits, force push. :+1:

johannsa:
Done.

luceos:
Commenting from mobile here. Two remarks there's a pluck to value rewrite that most likely should have been a separate pr. Additionally we can use the null coelescence operator when getting the description in the accessor.

johannsa:
Hi, @luceos!
1. I think the `pluck` to `value` affects this PR as that line is on a function used here and there has been an API change in Laravel and `pluck` does not return a single value anymore. And potentially this change in Laravel afects to other parts of the Flarum's codebase.
2. Null coalescing operator could be used, I agree. I tend to mix ternary and null coalescing operators, choosing them depending on the weather... :-D j/k

js/forum/src/components/TagHero.js
```diff
@@ -7,7 +7,10 @@ export default class TagHero extends Component {
     app.setTitle(tag.name());
     let description = tag.description();
     if (!description) {
-      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() }).join('');
+      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() });
+      if (description.constructor === Array) {
```
johannsa: The function `translator.apply`, used by `translator.trans`, returns **sometimes** an array instead of an string. That's the reason behind this check.

js/forum/src/components/TagHero.js
```diff
@@ -4,6 +4,15 @@ export default class TagHero extends Component {
   view() {
     const tag = this.props.tag;
     const color = tag.color();
+    app.setTitle(tag.name());
+    let description = tag.description();
+    if (!description) {
+      description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() });
+      if (description.constructor === Array) {
+        description = description.join('');
+      }
```
tobscure: Instead of this check, please use the `extractText` util. At the top of the file:

```js
import extractText from 'flarum/utils/extractText';
```

Then:

```js
description = app.translator.trans('flarum-tags.forum.meta_description.discussions_tagged_text', { tag: tag.name() });
app.setDescription(extractText(description));
```

franzliedke:
@johannsa Thank you for your effort on these!

Can you take care of the requested changes and rebase, please?

src/Forum/Controller/TagController.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    protected function getView(Request $request)
+    {
+        $view = parent::getView($request);
+        $id = $this->tags->getIdForSlug(array_get($request->getQueryParams(), 'slug'));
+        $actor = $request->getAttribute('actor');
+        $tag = $this->tags->findOrFail($id, $actor);
+
+        $description = $tag->description;
+        if ($description == null) {
```
luceos: This should have an exact match `===`.

PR #50, opened 2018-05-24
Use page component for tags page

clarkwinkelmann:
For some mysterious reasons the tags page wasn't using the Page component before.

I also fixed the history stack that was messed up when loading the tags page (icon passed as the title, resulting in `[Object object]` as the title of the back button...

This PR requires the new translation added in https://github.com/flarum/flarum-ext-english/pull/123


tobscure:
Nice catch, thanks

PR #54, opened 2018-07-15
Fix for new discussion button width

TurboProgramming:
See https://github.com/flarum/core/issues/1487 for details.

### DIFF
--- a/less/forum.less
+++ b/less/forum.less
@@ -71,6 +71,6 @@
     }
   }
   .TagsPage-nav .item-newDiscussion .Button {
-    width: auto;
+    width: 100%;
   }
 }


tobscure:
I don't think this fix will work in the context of a fix for https://github.com/flarum/core/issues/1486. The two issues are closely related so I think it will be easier to just have one PR that tackles both of them.

TurboProgramming:
@tobscure I completely agree with your decision and it seems smart, but until we get the redesign for the tags page, it may be better if we fix this temporarily as we already have the fix.

We also don't have any point in having separate lines of code for the same button in different pages. These are the exact same buttons but one operates with `auto` and the other operates with `100%`.

tobscure:
A fix for flarum/core#1486 and this is slated for beta 8.

PR #55, opened 2018-07-17
Replacement of fa-large with fa-tags for future redesign

TurboProgramming:
As we're going to see a redesign on the tags page, it would be better to see a replacement of the th-large icon with the tags icon for it to represent the page better and be general.

https://fontawesome.com/icons/tags?style=solid


luceos:
It would be smart to decide whether a pr like this is useful or not by getting feedback on our community forums beforehand, confirmation by a core developer would be best.

TurboProgramming:
@luceos I wanted to discuss it here, let me open a topic at the community.

PR #56, opened 2018-07-19
Add support for "own" permissions

tobscure:
Since tags executes its own catch-all permission logic for discussions,
we need to write in a special case for the new "own" permissions which
are not configurable per-tag.

See flarum/core#1512


tobscure:
Closing as per https://github.com/flarum/core/pull/1512#issuecomment-406456590

PR #58, opened 2018-10-05
Do not hide tags from Following (subscriptions)

datitisev:
**Fixes flarum/core#1232**

- ~~Do not hide discussions with hidden tag if `SubscriptionGambit` is used~~
  - ~~Note: No error is thrown if flarum/flarum-ext-subscriptions is not enabled~~
- Do not hide discussions with hidden tag if any gambit is used

- ~~Maybe we need a way to let extensions extend the list of gambits to ignore~~

**Screenshot**
image
image

**Confirmed**

- [ ] Frontend changes: tested on a local Flarum installation.
- [x] Backend changes: tests are green (run `php vendor/bin/phpunit`).

### DIFF
--- a/src/Listener/FilterDiscussionListByTags.php
+++ b/src/Listener/FilterDiscussionListByTags.php
@@ -13,6 +13,7 @@ namespace Flarum\Tags\Listener;
 
 use Flarum\Discussion\Event\Searching;
 use Flarum\Event\ConfigureDiscussionGambits;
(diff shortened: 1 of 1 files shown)


tobscure:
Maybe we could just say if *any* gambit is used then the tag shouldn't be hidden

datitisev:
@tobscure Hm, good idea, that way there's no need for extensions to add their own.

PR #59, opened 2018-11-04
Use ItemList for EditTagModal fields

clarkwinkelmann:
Similar to flarum/core#1593 and flarum/core#1625

- [x] Tested locally


luceos:
Fixes https://github.com/flarum/core/issues/1177

luceos:
@flarum/core this would simplify implementation for paywall and later on also affecting patronage (for patreon integration).

PR #62, opened 2019-06-02
Update code to work with latest html5sortable

datitisev:
Fixes flarum/core#1786

- Updated `import` statement, uses ES6 dist file
- Use `forEach` to add `sortupdate` event listener
- Check primary -> secondary by looking at element constructors
    - `ol` => `HTMLOListElement`
    - `ul` => `HTMLUListElement`
- Get data attribute through vanilla JS

### DIFF
--- a/js/src/admin/components/TagsPage.js
+++ b/js/src/admin/components/TagsPage.js
@@ -80,72 +80,72 @@ export default class TagsPage extends Page {
   }
 
   config() {
+    const cb = el => el.addEventListener('sortupdate', (e) => {
+      // If we've moved a tag from 'primary' to 'secondary', then we'll update
+      // its attributes in our local store so that when we redraw the change
+      // will be made.
+      if (e.detail.origin.container instanceof HTMLOListElement && e.detail.destination.container instanceof HTMLUListElement) {
+        app.store.getById('tags', e.detail.item.getAttribute('data-id')).pushData({
+          attributes: {
+            position: null,
+            isChild: false
+          },
+          relationships: {parent: null}
+        });
+      }
+
+      // Construct an array of primary tag IDs and their children, in the same
+      // order that they have been arranged in.
+      const order = this.$('.TagList--primary > li')
+        .map(function() {
+          return {
+            id: $(this).data('id'),
+            children: $(this).find('li')
+              .map(function() {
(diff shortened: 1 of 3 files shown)


js/src/admin/components/TagsPage.js
```diff
@@ -80,14 +80,15 @@ export default class TagsPage extends Page {
   }
 
   config() {
-    this.$('ol, ul')
-      .sortable({connectWith: 'primary'})
-      .on('sortupdate', (e, ui) => {
+    sortable(this.$('ol, ul'), {
+      acceptFrom: 'ol,ul'
+    }).forEach(el =>
+      el.addEventListener('sortupdate', (e) => {
```
franzliedke: Shouldn't we detach event listeners somewhere, to prevent leaking memory?

datitisev: Hm, I hadn't thought about that.

After quickly reading https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Memory_issues, maybe making a constant variable at the beginning of `config` set to the function and then use `forEach` and call the function, would that prevent memory leaks ? Not sure if it would be considered a static function...

```js
config() {
    const cb = el => ...;

   sortable(this.$('ol, ul'), {
      acceptFrom: 'ol,ul'
    }).forEach(cb);
}
```

franzliedke: Uhm, never mind, my comment about detaching was stupid. Maybe I should stick to my good old backend. :grin: 

But you're right, your suggestion would be an improvement in terms of unnecessary function creation.

franzliedke:
Other than that, thanks for fixing this. :+1:

datitisev:
Done :+1:

franzliedke:
What about extracting a method instead of a huge inline function? :grin: 

Then your `cb` var would just hold `this.extractedfunc.bind(this)`.

datitisev:
I guess that can be done. I don't see the point of it, but maybe for extending capabilities... for some reason 🤔

franzliedke:
@datitisev The point is complexity. Separating the code that sets up the event bindings, from the logic that happens when the event is triggered, for example. :wink: 

Great job, thanks again!

PR #63, opened 2019-07-07
Fix default value for useColor overriding falsy values

datitisev:
This would make the 'useColor' setting irrelevant as it would always be true.
Error is noticeable in the discussion list when the icons are the same color as the tag background.

Caused by 12acf03f4bfbd5fb45173aebc7a6966be1c09ce9.

`false` does not soft equal `null`, so we can use `!= null` to check if it is provided without having a whole if block.

### DIFF
--- a/js/src/common/helpers/tagIcon.js
+++ b/js/src/common/helpers/tagIcon.js
@@ -1,6 +1,6 @@
 export default function tagIcon(tag, attrs = {}, settings = {}) {
   const hasIcon = tag && tag.icon();
-  const useColor = settings.useColor || true;
+  const useColor = settings.useColor != null ? settings.useColor : true;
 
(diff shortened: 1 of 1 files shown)


js/src/common/helpers/tagIcon.js
```diff
@@ -1,6 +1,6 @@
 export default function tagIcon(tag, attrs = {}, settings = {}) {
   const hasIcon = tag && tag.icon();
-  const useColor = settings.useColor || true;
+  const useColor = settings.useColor != null ? settings.useColor : true;
```
franzliedke: Out of interest, does this work, too?

~~~js
const {useColor = true} = settings;
~~~

datitisev: Looks like it. That would also work in the parameters (`, { useColor = true } = {}`), but probably wouldn't look great.

Committed change :+1:.

franzliedke: Great, thanks. Especially for debugging and the quick fix!

PR #65, opened 2019-08-14
prevent lingering permissions when unrestricting

luceos:
Whenever a tag is no longer being restricted, a permission entry in group_permission lingers around for `viewDiscussions`. This PR removes all permissions configured after you stop restricting them.

### DIFF
--- a/src/Command/EditTagHandler.php
+++ b/src/Command/EditTagHandler.php
@@ -9,7 +9,6 @@
 
 namespace Flarum\Tags\Command;
 
-use Flarum\Group\Permission;
 use Flarum\Tags\Event\TagWillBeSaved;
 use Flarum\Tags\TagRepository;
 use Flarum\Tags\TagValidator;
@@ -88,7 +87,7 @@ class EditTagHandler
         $this->validator->assertValid($tag->getDirty());
 
         if ($tag->isDirty('is_restricted') && ! $tag->is_restricted) {
-            Permission::where('permission', 'like', 'tag'.$tag->id.'.%')->delete();
+            $tag->deletePermissions();
         }
 
         $tag->save();
--- a/src/Command/EditTagHandler.php
+++ b/src/Command/EditTagHandler.php
@@ -9,6 +9,7 @@
 
 namespace Flarum\Tags\Command;
 
+use Flarum\Group\Permission;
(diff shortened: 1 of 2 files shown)


src/Command/EditTagHandler.php
```diff
@@ -88,6 +89,10 @@ public function handle(EditTag $command)
 
         $this->validator->assertValid($tag->getDirty());
 
+        if ($tag->isDirty('is_restricted') && ! $tag->is_restricted) {
+            Permission::where('permission', 'like', 'tag'.$tag->id.'.%')->delete();
+        }
```
franzliedke: I don't know if we do stuff like that anywhere else, but queries like this and the one below should probably be wrapped in a transaction, right?

clarkwinkelmann: I don't know if we have transactions anywhere else.

At minimum I'd say delete the related permissions after saving the tag. This way, should an error happen, it will just result to the old behaviour, having unneeded permissions in DB. The current code could end up in a situation where the restriction isn't removed but some permissions have been, which would not be obvious to the forum admin.

franzliedke: Indeed, it should at least be moved down.

franzliedke:
I moved the permission deletion to a `saved` event in the model itself. How do you guys like this approach?

src/Tag.php
```diff
@@ -164,6 +170,20 @@ public function scopeWithStateFor(Builder $query, User $user)
...
+     *
+     * @return bool
+     */
+    public function wasUnrestricted()
+    {
+        if ($this->is_restricted) {
+            return false;
+        }
+
+        return array_key_exists('is_restricted', $this->getChanges());
```
luceos: Why not `$this->getDirty('is_restricted')`?

luceos:
See my comment, other than that, looking good!

src/Tag.php
franzliedke: Because we now delete the permission *after* saving the tag. At that point, the dirty state is already reset, but thankfully Laravel exposes previously saved model changes with `getChanges()`.

franzliedke: But thanks for bringing it up. This made me go through Laravel's code again and find the `wasChanged()` helper method. That makes the code even more concise. :grin:

PR #67, opened 2019-11-06
Infinite nesting of tags

luceos:
Added ability for infinitely nested tags.
Admin can fully cope with any nested level.
Frontend has minimal changes, it works, but visual display is funky.

Reviewers should focus on:

- [ ] frontend improvements
- [ ] do we want this

Created because we need this as an intermediate solution for @Bokt. There's no obligation to merge this, but doing so would reduce our overhead.

### DIFF
--- a/js/src/admin/components/TagsPage.js
+++ b/js/src/admin/components/TagsPage.js
@@ -20,7 +20,7 @@ function tagItem(tag) {
           onclick: () => app.modal.show(new EditTagModal({tag}))
         })}
       </div>
-      {!tag.isChild() && tag.position() !== null ? (
+      {tag.position() !== null ? (
         <ol className="TagListItem-children">
           {sortTags(app.store.all('tags'))
             .filter(child => child.parent() === tag)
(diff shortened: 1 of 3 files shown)


the-turk:
Is this still a thing?

luceos:
@the-turk there has been no discussion yet whether this is something we want/need. So I am assuming this PR will be pending for quite a while before we will take action. Until then it can be used for inspiration if people need it.

askvortsov1:
I've thought quite a bit about this, and here's my stance. I don't think we should do infinite tags. There's a lot of UI issues/implications, but more importantly, I don't think it's that consistent with Flarum's philosophy. Our treatment of secondary tags as first-class citizens, plus the ability to categorize a discussion under multiple tags, means that categorization needs are more than met. 

If we want to deepen the primary tag options, I would be in favor of offering 3-level tags. This should be more than enough for well-designed forums. If it's not enough, we're likely dealing with an extreme edge case (ala bokt), at which point, that forum:
- A: maintaining their own fork
- B: using some other kind of multi-tenant/other split of forum into several 'sub forums'

would make more sense than extremely nested categories.

luceos:
I fully agree. Even with this PR originating from my work with Bokt and it's multi nested categories, it just doesn't feel right to push this into our own extension.

PR #70, opened 2019-12-09
Allow extensions to place items between tags link and tags list

andreasjacobsen93:
This PR sets the priority of the tags list to -12, to allow placing a menu link on -11 between the link and the list.

Issue here: https://github.com/flarum/core/issues/1923

### DIFF
--- a/js/src/forum/addTagList.js
+++ b/js/src/forum/addTagList.js
@@ -32,7 +32,7 @@ export default function() {
         active = currentTag.parent() === tag;
       }
 
-      items.add('tag' + tag.id(), TagLinkButton.component({tag, params, active}), -12);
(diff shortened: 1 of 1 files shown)


datitisev:
We may want to make it -15, just to use multiples of 5.
Also - perhaps have a space between the tags elements as well? Like between the separator, tags list, and more tags link, in case an extension wants to add something there? Thoughts welcome.

luceos:
How about?

- separator 12
- tags 14
- more 16

andreasjacobsen93:
Updated PR to set separator at -12, tags at -14 and more at -16.

luceos:
Thank you @andreasjacobsen93 👍

PR #72, opened 2020-02-13
Added title and description meta tags

askvortsov1:
Modified code to set the meta "title" attribute based on the current page's tag, or to "Tags" (translatable) for pages related to the tags extension.

### DIFF
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -31,8 +31,6 @@ export default function() {
 
     if (tag) {
       app.setTitle(tag.name());
-    } else {
-      app.setTitle("");
     }
   });
 
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -23,8 +23,16 @@ export default function() {
   extend(IndexPage.prototype, 'view', function(vdom) {
     const tag = this.currentTag();
 
+    if (tag) vdom.attrs.className += ' IndexPage--tag'+tag.id();
+  });
+
+  extend(IndexPage.prototype, 'config', function () {
+    const tag = this.currentTag();
+
     if (tag) {
-      vdom.attrs.className += ' IndexPage--tag'+tag.id();
+      app.setTitle(tag.name());
+    } else {
+      app.setTitle("");
     }
   });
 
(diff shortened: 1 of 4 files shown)


askvortsov1:
With the changes made in f1c65ae, this fully covers requirements of #69 and #48, which can probably both be closed if this is good to merge.

luceos:
This is a very decent PR @askvortsov1 well done!

js/src/forum/addTagFilter.js
```diff
@@ -44,8 +52,12 @@ export default function() {
...
   // Add a parameter for the IndexPage to pass on to the DiscussionList that
   // will let us filter discussions by tag.
-  extend(IndexPage.prototype, 'params', function(params) {
-    params.tags = m.route.param('tags');
+  extend(IndexPage.prototype, 'params', function (params) {
+    if (m.route.param('tags')) {
+      params.tags = m.route.param('tags');
+    } else {
+      params.tags = "";
+    }
```
franzliedke: Why is this change necessary exactly? Just making sure we have a string here instead of `undefined` if there's no tag? If so, do we want that, or do we want to remove the param?

js/src/forum/addTagFilter.js
```diff
@@ -23,8 +23,16 @@ export default function() {
...
+  });
+
+  extend(IndexPage.prototype, 'config', function () {
+    const tag = this.currentTag();
+
     if (tag) {
-      vdom.attrs.className += ' IndexPage--tag'+tag.id();
+      app.setTitle(tag.name());
+    } else {
+      app.setTitle("");
```
franzliedke: Single quotes, please. :speak_no_evil: 

Sorry, we need a linter or a formatter. Will make that a priority.

franzliedke: Actually... does this mean that the title will suddenly be empty when moving from the unfiltered "All discussions" page to a tag page and then back to "All discussions"?

askvortsov1: Not sure why I had this, but after further testing it seems unnecessary.

js/src/forum/addTagFilter.js
askvortsov1: As with above, don't remember why it was added, but seems to work fine without it so I removed it.

askvortsov1:
Addressed

franzliedke:
Thank you!

askvortsov1:
@franzliedke the accompanying https://github.com/flarum/lang-english/pull/154 still needs to be merged

franzliedke:
I know. I will change the keys, though.

PR #74, opened 2020-02-19
Fixed restricted tags having enabled "Start Discussion" button

askvortsov1:
Fixes https://github.com/flarum/core/issues/1696

### DIFF
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -29,7 +29,7 @@ export default function() {
   });
 
   // If currently viewing a tag, restyle the 'new discussion' button to use
-  // the tag's color.
+  // the tag's color, and disable if the user isn't allowed to edit.
   extend(IndexPage.prototype, 'sidebarItems', function(items) {
     const tag = this.currentTag();
 
@@ -39,6 +39,9 @@ export default function() {
       if (color) {
         items.get('newDiscussion').props.style = {backgroundColor: color};
       }
+    
+      items.get('newDiscussion').props.disabled = !tag.canStartDiscussion();
+    
     }
   });
 
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -35,12 +35,14 @@ export default function() {
 
     if (tag) {
       const color = tag.color();
+      const canStartDiscussion = tag.canStartDiscussion();
 
       if (color) {
         items.get('newDiscussion').props.style = {backgroundColor: color};
       }
     
-      items.get('newDiscussion').props.disabled = !tag.canStartDiscussion();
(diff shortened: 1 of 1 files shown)


js/src/forum/addTagFilter.js
```diff
@@ -29,16 +29,21 @@ export default function() {
...
 
     if (tag) {
       const color = tag.color();
+      const canStartDiscussion = tag.canStartDiscussion();
 
       if (color) {
         items.get('newDiscussion').props.style = {backgroundColor: color};
       }
+    
+      items.get('newDiscussion').props.disabled = !canStartDiscussion;
```
franzliedke, line 44: I took a look at [how this was implemented in core](https://github.com/flarum/core/blob/7794546845c503c39671de0f465963dc566c972c/js/src/forum/utils/DiscussionControls.js#L68) for the "Reply" / "Can't reply" button (inside a discussion).

I noticed that we are not explicitly disabling that button there, but instead only styling it in such a way. I was going to change that, because the `disabled` attribute signals semantics, but unfortunately it also removes the button from the tab order. That [is an accessibility problem](https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes), so let's not do that. Changing the label is great, please add the `disabled` class instead of the attribute!

franzliedke: Hmm, apparently we do disable the "Start a discussion" completely when the whole forum isn't writable. :thinking:

askvortsov1: I would propose that we keep the 'start discussion' button consistent for now (e.g. use the disabled prop), and open a second PR changing that behavior to use the 'disabled' class.

tankerkiller125: @franzliedke In regards to the accessibility issue, if the button isn't being displayed in the first place because user doesn't have permissions to post there would it even make sense to include it in screen readers/tab hints? It seems like that might create even more confusion especially for people who rely on screen readers as to them the button would exist but they would be unable to use it and at least in our current system there is nothing to tell them why it's not working.

allroundernaman:
Hey is there any way that I can restrict the use of tags for normal members? Like I made a tag of News and Announcements where I don't want people to make new discussions, I simply want it to be read-only.

PeopleInside:
Hi @allroundernaman, this should be done on the permission page where you can add a label and give the desired permission.
For questions like this you can use the Flarum community https://discuss.flarum.org/

allroundernaman:
Okay sure. But I can't see additional permissions on the administration? Is there is something extra I need to install.

allroundernaman:
Thanks I got that. 
Regards,

allroundernaman:
Thanks but I think I have got that correct.
Regards,

PR #75, opened 2020-02-19
Performance improvements

imorland:
Combined effort from several members of the @giffgaff team, contributing our efforts back here.

Generally, replacing `whereExists/whereNotExists` with `whereIn/whereNotIn`

### DIFF
--- a/src/Access/DiscussionPolicy.php
+++ b/src/Access/DiscussionPolicy.php
@@ -76,11 +76,10 @@ class DiscussionPolicy extends AbstractPolicy
     public function find(User $actor, Builder $query)
     {
         // Hide discussions which have tags that the user is not allowed to see.
-        $query->whereNotExists(function ($query) use ($actor) {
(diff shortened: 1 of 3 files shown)


datitisev:
What is that `getDiscussionIdsWithHiddenSubscriptionChildTagsQry` code...? That looks like it should not be part of a core PR (looks like integation with fof/follow-tags?).

imorland:
Yes you're correct. Apologies, I will get that updated shortly

franzliedke:
Ah, I think I understand. The `IN` variant has a subquery that can be used for *all* rows of the outer query, whereas the `EXISTS` one would trigger a subquery for each outer row. At least that's my understanding.

Thank you, Giffgaff!

imorland:
Sorry for the late reply here. Yes, you are exactly right @franzliedke, that is our understanding too

franzliedke:
No worries, it's not like I gave you a lot of time. :laughing:

PR #79, opened 2020-04-05
Increase Tags Visibility Scoping Extendability

askvortsov1:
Fixes https://github.com/flarum/core/issues/2116.

Right now, if a user doesn't have the viewDiscussions permission for a tag, they CAN NOT see discussions in the tag, and there is no ability for extensions to override this. This PR introduces a scopeModelVisibility event dispatch, which allows extensions to override this.

An example of a use case can be found here:
https://github.com/askvortsov1/flarum-help-tags/issues/1

### DIFF
--- a/src/Access/DiscussionPolicy.php
+++ b/src/Access/DiscussionPolicy.php
@@ -27,6 +27,11 @@ class DiscussionPolicy extends AbstractPolicy
     protected $model = Discussion::class;
 
     /**
+     * @var Dispatcher
+     */
+    protected $events;
+
+    /**
      * @var SettingsRepositoryInterface
      */
     protected $settings;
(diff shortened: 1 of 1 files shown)


src/Access/DiscussionPolicy.php
```diff
@@ -75,12 +76,18 @@ public function can(User $actor, $ability, Discussion $discussion)
...
     {
-        // Hide discussions which have tags that the user is not allowed to see.
-        $query->whereNotIn('discussions.id', function ($query) use ($actor) {
-            return $query->select('discussion_id')
-                ->from('discussion_tag')
-                ->whereIn('tag_id', Tag::getIdsWhereCannot($actor, 'viewDiscussions'));
-        });
+        // Hide discussions which have tags that the user is not allowed to see, unless an extension overrides this.
+        $query
+            ->whereNotIn('discussions.id', function ($query) use ($actor) {
```
clarkwinkelmann, line 81: I would suggest wrapping the `whereNotIn()->orWhere()` inside another `->where(function($query) {})` to prevent introducing ORs in the query outside of condition groups

askvortsov1: Good point, done!

PR #80, opened 2020-04-23
Trying to Implement defaultSort for Tags

KellyBrakke:
Based on post #328.

### DIFF
--- a/js/src/admin/components/EditTagModal.js
+++ b/js/src/admin/components/EditTagModal.js
@@ -21,6 +21,7 @@ export default class EditTagModal extends Modal {
     this.color = m.prop(this.tag.color() || '');
     this.icon = m.prop(this.tag.icon() || '');
     this.isHidden = m.prop(this.tag.isHidden() || false);
+    this.defaultSort = m.prop(this.tag.defaultSort() || '');
   }
 
   className() {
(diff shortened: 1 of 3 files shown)


src/Command/EditTagHandler.php
```diff
@@ -82,6 +82,10 @@ public function handle(EditTag $command)
             $tag->is_restricted = (bool) $attributes['isRestricted'];
         }
 
+        if (isset($attributes['defaultSort'])) {
+            $tag->is_restricted = $attributes['defaultSort'];
```
askvortsov1, line 86: Do you mean $tag->default_sort?

src/Command/CreateTagHandler.php
```diff
@@ -47,7 +47,8 @@ public function handle(CreateTag $command)
             array_get($data, 'attributes.description'),
             array_get($data, 'attributes.color'),
             array_get($data, 'attributes.icon'),
-            array_get($data, 'attributes.isHidden')
+            array_get($data, 'attributes.isHidden'),
+            array_get($data, 'attributes.defaultSort')
```
askvortsov1, line 51: You're going to want to edit Tag::build to accomodate the extra parameter.

js/src/admin/components/EditTagModal.js
```diff
@@ -21,6 +21,7 @@ export default class EditTagModal extends Modal {
     this.color = m.prop(this.tag.color() || '');
     this.icon = m.prop(this.tag.icon() || '');
     this.isHidden = m.prop(this.tag.isHidden() || false);
+    this.defaultSort = m.prop(this.tag.defaultSort() || '');
```
askvortsov1, line 24: This should probably default to 50? (as per the migration)

PR #81, opened 2020-05-12
Remove !important because it's not needed

BigMinnie:
Remove !important because it's not needed and additional changing force you to hack the theme with animations that does not work on all browsers.

### DIFF
--- a/less/common/TagLabel.less
+++ b/less/common/TagLabel.less
@@ -16,22 +16,22 @@
 
   &.colored {
     .TagLabel-text {
-      color: @body-bg !important;
+      color: @body-bg;
     }
   }
 
   .DiscussionHero .TagsLabel & {
     background: transparent;
-    border-radius: @border-radius !important;
+    border-radius: @border-radius;
     font-size: 14px;
 
     &.colored {
       margin-right: 5px;
-      background: @body-bg !important;
+      background: @body-bg;
       color: @muted-color;
 
       .TagLabel-text {
-        color: inherit !important;
+        color: inherit;
       }
     }
   }


BigMinnie:
Hey @askvortsov1, sorry for the ping but I saw that you approved the changes so I'm just wondering if you are planing to push it into the final build?

tankerkiller125:
@BigMinnie we require at least two devs to approve before merges, further at least one of those devs needs to test locally. I personally don't have time to review this tonight (my time) but if it's not reviewed by anyone else by the time I get off work tomorrow I'll see if I can take a look. No promises though.

BigMinnie:
Yeah forgot that I tested it with my style on so when I removed !important my style was taking it over. I found it out why you guys used !important. Someone inserted CSS style into the elements itself and this is probably done by this extension. But sadly my knowledge is just in CSS so I do not know javascript and php. 
Screenshot_1

datitisev:
@BigMinnie that is done through JavaScript for the tag colors. We could actually change that so they are [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) instead and have the color & background color properties in the less files.

BigMinnie:
@datitisev I don't know why they are even there? Background color is already overwritten in less file with !important and color could also be there. So elements would not need any of those styles. And if everything would move into the css file there would be no need for any !important.

datitisev:
@BigMinnie If I remember correctly, `style` attribute styles overwrite CSS `!important` ones in external files. The background color !important(s) are probably for the default color in light & dark mode.

Those colors cannot be moved into the CSS files because they are user-specific (tags). However, they _can_ be made into CSS variables and those can be used instead (see above comment).

Actually - we could probably do something with custom LESS for tags? Not sure - setting variable and then looping values / adding custom classes in JS & LESS 🤔

BigMinnie:
@datitisev what I meant is that there is no reason that there is a background color inside elements style if you are trying to overwrite it with the same extension in CSS. But yes, this variables would the best solution.

PR #82, opened 2020-05-17
Fix buggy drag and drop

KyrneDev:
The current Tags drag and drop in admin is really buggy and hard to use. It is almost impossible to use on mobile. This updates the sortable library to something more modern which makes it much, much easier to use (and has a nice little animation).

### Before: 
OldVersion

### Mobile (totally unusable):
Old Mobile

### New: 
NewVersion

### Mobile: 
New Mobile

### DIFF
--- a/js/src/admin/components/TagsPage.js
+++ b/js/src/admin/components/TagsPage.js
@@ -85,6 +85,7 @@ export default class TagsPage extends Page {
               group: 'tags',
               animation: 150,
               swapThreshold: 0.65,
+			  dragClass: 'sortable-dragging',
               ghostClass: 'sortable-placeholder',
               onSort: (e) => this.onSortUpdate(e)
           })
--- a/js/src/admin/components/TagsPage.js
+++ b/js/src/admin/components/TagsPage.js
@@ -85,7 +85,7 @@ export default class TagsPage extends Page {
               group: 'tags',
               animation: 150,
               swapThreshold: 0.65,
-	      dragClass: 'sortable-dragging',
+	            dragClass: 'sortable-dragging',
               ghostClass: 'sortable-placeholder',
               onSort: (e) => this.onSortUpdate(e)
           })
--- a/js/src/admin/components/TagsPage.js
+++ b/js/src/admin/components/TagsPage.js
@@ -85,7 +85,7 @@ export default class TagsPage extends Page {
               group: 'tags',
               animation: 150,
               swapThreshold: 0.65,
-			  dragClass: 'sortable-dragging',
+	      dragClass: 'sortable-dragging',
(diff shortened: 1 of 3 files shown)


KyrneDev:
Please note that I changed to Dragula as per @datitisev's request. The functionality on mobile remains the same, but there is no animation as shown above.

askvortsov1:
@KyrneDev @datitisev are there plans to make additional changes in animation to this, or is this ready for review?

datitisev:
Imma play around with it.

datitisev:
FYI - you removed the dist files instead of resetting them to a previous commit. That is fine, but you got me confused as to why the dist folder didn't exist.

js/src/admin/components/TagsPage.js
```diff
@@ -80,18 +80,24 @@ export default class TagsPage extends Page {
...
   config() {
-    sortable(this.$('ol, ul'), {
-      acceptFrom: 'ol,ul'
-    }).forEach(this.onSortUpdate.bind(this));
+      this.$('.TagList').get().map(e => {
+          sortable.create(e, {
+              group: 'tags',
+              animation: 150,
+              swapThreshold: 0.65,
+			  dragClass: 'sortable-dragging',
```
datitisev, line 88: Indent this correctly please 🙂

datitisev:
Oh, one thing I forgot to add. We probably want to expose sortablejs to window - so other extensions can take advantage of it or lazy load their own.

askvortsov1:
How would we do that? webpack/importing/exporting JS is still a minefield for me

datitisev:
With `expose-loader`. It'd be like [this](https://github.com/flarum/core/blob/3b1f5ca07b879e1ccc2466bac8fbf343d2359e75/js/src/common/index.js#L3), so `import 'expose-loader?sortable!sortablejs`.

clarkwinkelmann:
> @datitisev  We probably want to expose sortablejs to window - so other extensions can take advantage of it or lazy load their own

I'm not seeing a benefit here. If another extensions needs `sortable` but doesn't depend on the Tags extension, they will need to include the `sortable` library in their dist file anyway in case Tags isn't being used.

Unless we also have a way for extension to conditionally add javascript dependencies only when it's not already registered, there won't be any benefit from a bundle size perspective, and I'm not seeing any other possible benefit.

datitisev:
@clarkwinkelmann I mentioned lazy-loading because they can still request a JS file after building - it'll just need to be in the extension's assets folder or use an external CDN.

See https://github.com/FriendsOfFlarum/drafts/blob/aece8dfe3f0b294f17807258b229f16857a0957a/js/src/forum/components/ScheduleDraftModal.js for an example.

PR #83, opened 2020-05-20
Fix tag selection input on discussion modal going out of screen

askvortsov1:
Fix https://github.com/flarum/core/issues/2040
Fix https://github.com/flarum/core/issues/2030

Behavior before:
https://youtu.be/w1XLzFZFIqg

Behavior now:
https://youtu.be/ZkvKr60J-vk

### DIFF
--- a/js/src/forum/components/TagDiscussionModal.js
+++ b/js/src/forum/components/TagDiscussionModal.js
@@ -155,7 +155,7 @@ export default class TagDiscussionModal extends Modal {
         <div className="TagDiscussionModal-form">
           <div className="TagDiscussionModal-form-input">
             <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}
-              onclick={() => this.$('.TagsInput input').first().focus()}
+              onclick={() => this.$('.TagsInput input').focus()}
             >
               <span className="TagsInput-selected">
                 {this.selected.map(tag =>
@@ -167,7 +167,7 @@ export default class TagDiscussionModal extends Modal {
                   </span>
                 )}
               </span>
-              <input className="FormControl TagsInput-input"
+              <input className="FormControl"
                 placeholder={extractText(this.getInstruction(primaryCount, secondaryCount))}
                 value={this.filter()}
                 style={{ width: inputWidth + 'ch' }}
--- a/js/src/forum/components/TagDiscussionModal.js
+++ b/js/src/forum/components/TagDiscussionModal.js
@@ -155,7 +155,7 @@ export default class TagDiscussionModal extends Modal {
         <div className="TagDiscussionModal-form">
           <div className="TagDiscussionModal-form-input">
             <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}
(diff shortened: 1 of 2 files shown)


js/src/forum/components/TagDiscussionModal.js
```diff
@@ -148,11 +148,15 @@ export default class TagDiscussionModal extends Modal {
...
 
+    const inputWidth = Math.max(extractText(this.getInstruction(primaryCount, secondaryCount)).length, this.filter().length);
+
     return [
       <div className="Modal-body">
         <div className="TagDiscussionModal-form">
           <div className="TagDiscussionModal-form-input">
-            <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}>
+            <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}
+              onclick={() => [].forEach.call(document.getElementsByClassName("TagsInput-input"), (element) => element.focus())}
```
datitisev, line 158: For this, don't we want to use `this.$('.TagsInput input).get()` instead? (Can't remember if .get() returns array, but you get the point).

clarkwinkelmann: We basically commented the same thing :D

Can there be *multiple* inputs? There should be one one no?

datitisev: Hm, good point. Looping over each input element doesn't make much sense because, in the case there's more than one, it'd just focus on all and end on the last one.

js/src/forum/components/TagDiscussionModal.js
```diff
@@ -163,9 +167,10 @@ export default class TagDiscussionModal extends Modal {
                   </span>
                 )}
               </span>
-              <input className="FormControl"
+              <input className="FormControl TagsInput-input"
```
clarkwinkelmann, line 170: You could even remove the className now I think?

js/src/forum/components/TagDiscussionModal.js
```diff
@@ -148,11 +148,15 @@ export default class TagDiscussionModal extends Modal {
...
 
+    const inputWidth = Math.max(extractText(this.getInstruction(primaryCount, secondaryCount)).length, this.filter().length);
+
     return [
       <div className="Modal-body">
         <div className="TagDiscussionModal-form">
           <div className="TagDiscussionModal-form-input">
-            <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}>
+            <div className={'TagsInput FormControl ' + (this.focused ? 'focus' : '')}
+              onclick={() => this.$('.TagsInput input').first().focus()}
```
clarkwinkelmann, line 158: I don't think `.first().focus()` is even necessary here? Calling `.focus()` directly should focus all fields, but we only expect one anyway.

datitisev: What if another extension adds a field though? Then it will focus on the last of the fields (I think) or cause confusing behavior.

askvortsov1: When I tested this, calling .focus() on a group ran it on the first field, so the behavior is the same.

PR #84, opened 2020-05-21
Remove unneeded withState when loading tags

askvortsov1:
Part of https://github.com/flarum/core/issues/2177.

Assembling withState for tags uses at least one query per-tag, which scales up to a LOT with many tags as per testing with @luceos. If we eliminate this unneeded strain, performance is increased.


datitisev:
The state is used for `marked_as_read_at` and `is_hidden` - though they don't seem to be used anywhere (there's a separate `is_hidden` for the tags themselves).

This state is used in fof/follow-tags for `subscription` - what would be a good way of including the state without repeating the Tag query there? 🤔

PR #85, opened 2020-05-31
Add expose loader and expose sortable to window

askvortsov1:
Can we use expose loader in core to expose expose loader with expose loader?

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -2540,6 +2540,11 @@
         "homedir-polyfill": "^1.0.1"
       }
     },
+    "expose-loader": {
+      "version": "0.7.5",
+      "resolved": "https://registry.npmjs.org/expose-loader/-/expose-loader-0.7.5.tgz",
+      "integrity": "sha512-iPowgKUZkTPX5PznYsmifVj9Bob0w2wTHVkt/eYNPSzyebkUgIedmskf/kcfEIWpiWjg3JRjnW+a17XypySMuw=="
+    },
     "extend-shallow": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/extend-shallow/-/extend-shallow-3.0.2.tgz",
@@ -2770,7 +2775,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -2788,11 +2794,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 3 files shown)


clarkwinkelmann:
What does this solves?

I think it's problematic for extensions to expose global methods or libraries as you can't expect the other extension to be enabled and there could be version conflicts if multiple extensions do it.

But I do think it would be great if Flarum included a global library to make sortable fields as it's a recurring thing extensions needs. But then this would need to be moved to core.

askvortsov1:
Sorry for the lack of context, this was prompted by https://github.com/flarum/tags/pull/82#issuecomment-636490163. The 'expecting extensions to be enabled' issue should hopefully eventually be solved in https://github.com/flarum/core/pull/2188, but I see your point. However, if we add all that stuff to core, it can result in a bloated core? I think that we can probably give bundled extensions a bit of special treatment here.

datitisev:
Not sure where we want to discuss this, but I added why I think this is a good idea & useful @ https://github.com/flarum/tags/pull/82#issuecomment-636504713.

clarkwinkelmann:
I don't think it makes sense to merge this PR at this time.

It would be best to open a dedicated issue about extensions loading common JS dependencies, deciding on a best practice, then implement that best practice in the Tags extension.

If we go for conditional dependency loading, then Tags must also consider that the dependency might have already been loaded by another core or community extension.

datitisev:
The conditional dependency loading wouldn't be in core extensions. It would be in third party extensions that decide to implement it. Merging this PR will not have any effect and will keep everything the same unless extensions implement the conditional lazy loading. 

There's no downside to merging this PR as far as I can see.

clarkwinkelmann:
I feel like making `sortable` available globally could encourage bad practice from extension developers expecting it to be available anytime, not realizing it's being provided by the Tags extension.

I'm pretty sure this exact issue played out with an early version of Mason/Masquerade or another similar extension, but I'm unable to find traces of the issue.

IMO it's not offering us any benefit as Flarum developers, and might introduce headaches later on depending what community extensions decide to do with this new "feature". If we want this library to be global, I'd say we move it to core and officially announce it to third-party developers. Otherwise wait until we have a best practice to suggest to developers regarding often-used javascript libraries and implement it in Tags the same way we suggest third-party developers do.

I think we need an issue to discuss this further, in particular in regard to library versions. Can this mess up if extensions need a different version of the library? I remember jQuery support was removed at some point, so extensions might still be using another version. Can there be conflicts?

A PR probably isn't the right place to talk about those use cases and edge cases.

datitisev:
> Can this mess up if extensions need a different version of the library?

No, they can import their own and use that instead. But it'll have to be bundled - not exported to window, as that'd replace the other one.

> IMO it's not offering us any benefit as Flarum developers, and might introduce headaches later on depending what community extensions decide to do with this new "feature".

It benefits the user & decreases bundle size if extensions take advantage of this.

> If we want this library to be global, I'd say we move it to core

Then the library could be provided with nobody using it. It's a good chunk of size that matters.

> I'm pretty sure this exact issue played out with an early version of Mason/Masquerade or another similar extension

I think this was html5sortable, maybe? Or whatever we used before it, not sure.

PR #86, opened 2020-06-16
Firing 2 new events: TagWillBeCreated & TagWillBeDeleted

iPurpl3x:
Hi, as mentioned in my [post on the forum](https://discuss.flarum.org/d/24212-add-data-to-a-tag-object-while-creating), I need these two events for an extension that I am writing. And I really think it could be useful to have these anyway!

_Question: Can these changes be released before the next beta release of flarum/core or will I need to wait for beta 14 and in the meantime use my own repo?_

### DIFF
--- a/src/Command/CreateTagHandler.php
+++ b/src/Command/CreateTagHandler.php
@@ -9,7 +9,7 @@
 
 namespace Flarum\Tags\Command;
 
-use Flarum\Tags\Event\TagWillBeCreated;
+use Flarum\Tags\Event\Creating;
 use Flarum\Tags\Tag;
 use Flarum\Tags\TagValidator;
 use Flarum\User\AssertPermissionTrait;
@@ -66,7 +66,7 @@ class CreateTagHandler
             }
         }
 
-        event(new TagWillBeCreated($tag, $actor, $data));
+        event(new Creating($tag, $actor, $data));
 
         $this->validator->assertValid($tag->getAttributes());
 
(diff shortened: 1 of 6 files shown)


tankerkiller125:
I haven't gotten a good chance to review the code, but to answer your question: We only release these extensions with Core upgrades at the moment, so if these changes are approved before beta 14 they would be released to the general Flarum user in beta 14

datitisev:
We recently changed the names of events in flarum/core. I think these events should be called `Creating` and `Deleting`.

iPurpl3x:
OK @datitisev, I changed the names. 

Should I also change the existing `TagWillBeSaved` to `Saving` ? This will have breaking impacts, that's why I ask.

datitisev:
Oh, I didn't realize the other event used the old naming convention. Hm, I don't think we want to change it then - probs best to ignore my previous comment and revert the latest naming changes.

Or you can wait for someone else's input on this :).

askvortsov1:
Hmm I'd prefer not to introduce changes that are already outdated unless it's a necessity. Would you be willing to follow up with a second PR that adds a Saving event as a copy paste of TagWillBeSaved, mark TagWillBeSaved as deprecated, have both events fire, and have any internal listeners listen to Saving instead?

iPurpl3x:
Alright @askvortsov1, I'll do that when I have some time. I also thing that's the right way to go.

PR #92, opened 2020-08-14
Renamed TagWillBeSaved => Saving. Deprecated the old event.

iPurpl3x:
As discussed in #86, here is a PR for the Saving event that was previously named TagWillBeSaved but should be renamed to match the general convention.
Is the `@deprecated beta 14, removed beta 15.` comment that I made right @askvortsov1 ?

### DIFF
--- a/src/Command/EditTagHandler.php
+++ b/src/Command/EditTagHandler.php
@@ -10,6 +10,7 @@
 namespace Flarum\Tags\Command;
 
 use Flarum\Tags\Event\Saving;
+use Flarum\Tags\Event\TagWillBeSaved;
 use Flarum\Tags\TagRepository;
 use Flarum\Tags\TagValidator;
 use Flarum\User\AssertPermissionTrait;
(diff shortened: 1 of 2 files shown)


src/Event/TagWillBeSaved.php
```diff
@@ -12,6 +12,9 @@
 use Flarum\Tags\Tag;
 use Flarum\User\User;
 
+/**
+ * @deprecated beta 14, removed beta 15.
+ */
```
franzliedke, line 17: If we keep this, we would need to keep raising it as well.

Alternative (and my preference): Just remove the event. :wink:

franzliedke (changes requested):

src/Event/TagWillBeSaved.php
franzliedke: Pretty much what Clark said, sorry.

franzliedke: Option 1 was actually requested by @askvortsov1 in https://github.com/flarum/tags/pull/86#issuecomment-646032898 as well.

luceos:
Looking at the discussion this is what still has to be done:

- [ ] Keep firing the deprecated event
- [ ] Create a new issue to remove that deprecated event and assign it to the next release/milestone 

@iPurpl3x could you tackle this so that we can ship it with the upcoming release?

@clarkwinkelmann let's create an issue for that, I think we need to make this a convention anyhow. We can then handle this in a new PR across all core extensions.

PR #93, opened 2020-08-15
Mithril 2 update

askvortsov1:


js/src/admin/components/EditTagModal.js
```diff
@@ -60,31 +60,31 @@ export default class EditTagModal extends Modal {
 
     items.add('slug', <div className="Form-group">
       <label>{app.translator.trans('flarum-tags.admin.edit_tag.slug_label')}</label>
-      <input className="FormControl" value={this.slug()} oninput={m.withAttr('value', this.slug)}/>
+      <input className="FormControl" bidi={this.slug}/>
```
clarkwinkelmann, line 63: is the change from `oninput` to `onchange` voluntary? (as a consequence of using `bidi`)

Same for lines 68, 73 and 81.

js/src/admin/components/TagsPage.js
```diff
@@ -32,6 +32,12 @@ function tagItem(tag) {
 }
 
 export default class TagsPage extends Page {
+  oninit(vnode) {
+    super.oninit(vnode);
+
+    this.key = 0;
```
clarkwinkelmann, line 38: I'd suggest maybe moving the comment from the bottom of the file to here as to not confuse new readers.

I'd also suggest naming the variable differently, like `forcedRefreshKey` for clarity and making extra sure it won't be conflicting with any Mithril `key` feature at the component level.

js/src/forum/addTagList.js
```diff
@@ -32,7 +31,7 @@ export default function() {
         active = currentTag.parent() === tag;
       }
 
-      items.add('tag' + tag.id(), TagLinkButton.component({tag, params, active}), -14);
+      items.add('tag' + tag.id(), <TagLinkButton hello="there" model={tag} params={params} active={active} />, -14);
```
clarkwinkelmann, line 34: `hello="there"` ?

Also, wouldn't it be more compact to keep the `.component()` syntax here? Only the `tag` would need to change to `model: tag`.

js/src/forum/components/TagDiscussionModal.js
```diff
@@ -169,22 +169,17 @@ export default class TagDiscussionModal extends Modal {
               </span>
               <input className="FormControl"
                 placeholder={extractText(this.getInstruction(primaryCount, secondaryCount))}
-                value={this.filter()}
+                bidi={this.filter}
```
clarkwinkelmann, line 172: Also changes from `oninput` to `onchange`. I think this could lead to issues if a redraw is triggered from `onkeydown` as this could apply the previous value to the field?

js/src/forum/components/TagDiscussionModal.js
```diff
@@ -210,12 +205,12 @@ export default class TagDiscussionModal extends Modal {
                 <span className="SelectTagListItem-name">
                   {highlight(tag.name(), filter)}
                 </span>
-                {tag.description()
-                  ? (
+                {!!tag.description()
```
clarkwinkelmann, line 208: Doesn't this render `0` when false?

I'd prefer keeping the original syntax, eventually replacing `''` with `null`.

js/src/admin/components/EditTagModal.js
askvortsov1: We use bidi for <input> elements widely across core and other extensions, I guess if there was an issue, it would have showed up by now?

js/src/forum/components/TagDiscussionModal.js
askvortsov1: See above

js/src/admin/components/EditTagModal.js
clarkwinkelmann: It's not about an issue with `bidi` itself. Usually the use of `oninput` means we're trying to redraw the interface on each keypress.

I would say it's *rare* to need `oninput` and it would make sense to have a comment indicating why `oninput` is used. Here we unfortunately don't have much insight into the 5-year-old code, so it's difficult to say whether it was intentional or not :grimacing: 

In this case, I don't see any code that would require immediate update of `this.slug`, so I think it's fine to replace it with `onchange`/`bidi`.

js/src/forum/components/TagDiscussionModal.js
clarkwinkelmann: Here my worry was that `onkeydown` would trigger a redraw of the results, but the filtering would use the value from `this.filter`, which is only update on `onchange`. So this means unless you click away from the field, `onchange` wouldn't get triggered and the interface would always show the same results matching the filter value from the last `onchange` event.

js/src/admin/components/EditTagModal.js
franzliedke: Not good - this has definitely been done very inconsistently in the big PR in core.

franzliedke: Aaaaactually... it seems like [the code](https://github.com/tobyzerner/m.attrs.bidi/blob/master/bidi.js#L38-L40) sets both `oninput` and `onchange`.

askvortsov1: So what's the verdict?

clarkwinkelmann: Oh I didn't know that. Not sure where I got the idea that it would only set `onchange`. So the change from `oninput` to `bidi` will definitely have no functional impact.

askvortsov1: So are we fine with these changes then?

js/src/forum/components/TagDiscussionModal.js
clarkwinkelmann: Now that we found out that `bidi` does call `oninput`, this argument is moot and I don't think there's any actual issue with this.

clarkwinkelmann (approved):
Looking good code-wise

js/src/forum/components/TagHero.js
```diff
@@ -3,7 +3,7 @@ import tagIcon from '../../common/helpers/tagIcon';
 
 export default class TagHero extends Component {
   view() {
-    const tag = this.props.tag;
+    const tag = this.attrs.model;
```
franzliedke, line 6: Was this changed because `tag` is reserved in Mithril or something?

js/src/forum/components/TagsPage.js
```diff
@@ -36,18 +36,14 @@ export default class TagsPage extends Page {
...
                       <p className="TagTile-description">{tag.description()}</p>
                       {children
                         ? (
                           <div className="TagTile-children">
                             {children.map(child => [
-                              <a href={app.route.tag(child)} config={function(element, isInitialized) {
-                                if (isInitialized) return;
-                                $(element).on('click', e => e.stopPropagation());
-                                m.route.apply(this, arguments);
-                              }}>
```
franzliedke, line 50: I've seen a change like this before... why do we no longer need the propagation stuff? cc @clarkwinkelmann

js/src/forum/components/TagsPage.js
```diff
@@ -23,7 +23,7 @@ export default class TagsPage extends Page {
       <div className="TagsPage">
         {IndexPage.prototype.hero()}
         <div className="container">
-          <nav className="TagsPage-nav IndexPage-nav sideNav" config={IndexPage.prototype.affixSidebar}>
+          <nav className="TagsPage-nav IndexPage-nav sideNav">
```
franzliedke, line 26: No replacement for the `affixSidebar` stuff? :thinking:

js/src/forum/components/TagsPage.js
```diff
@@ -85,8 +81,8 @@ export default class TagsPage extends Page {
     );
   }
 
-  config(...args) {
-    super.config(...args);
+  oncreate(vnode) {
+    super.oncreate(vnode);
```
franzliedke, line 85: No `onupdate` stuff here?

js/src/forum/components/TagDiscussionModal.js
```diff
@@ -317,10 +312,8 @@ export default class TagDiscussionModal extends Modal {
         });
     }
 
-    if (this.props.onsubmit) this.props.onsubmit(tags);
+    if (this.attrs.onsubmit) this.attrs.onsubmit(tags);
 
-    app.modal.close();
-
-    m.redraw.strategy('none');
```
franzliedke, line 324: I am missing an `e.redraw = false` here. But this change is confusing anyway. Oversight from applying modal state changes to all bundled extensions?

js/src/forum/addTagList.js
```diff
@@ -32,7 +31,12 @@ export default function() {
         active = currentTag.parent() === tag;
       }
 
-      items.add('tag' + tag.id(), TagLinkButton.component({tag, params, active}), -14);
+      // tag.name() is passed here as children even though it isn't used directly
+      // because when we need to get the active child in SelectDropdown, we need to
+      // use its children to populate the dropdown. The problem here is that `view`
+      // on TagLinkButton is only called AFTER SelectDropdown, so no children are available
+      // for SelectDropdown to use at the time.
+      items.add('tag' + tag.id(), TagLinkButton.component({model: tag, params, active}, tag?.name()), -14);
```
franzliedke, line 39: Optional chaining on `tag?.name()`... is that supported? Isn't that only a TypeScript thing?

js/src/forum/components/TagHero.js
askvortsov1: Yes

js/src/forum/addTagList.js
datitisev: If that works it's because of a Babel plugin. It's not explicitly a TS thing, I think.

js/src/forum/components/TagsPage.js
askvortsov1: On the TagsPage, the navbar is at the top, and is absolutely unaffixed. I believe I removed this because it didn't actually do anything (and that method was removed from IndexPage in core)

js/src/forum/components/TagsPage.js
askvortsov1: Page title and title count doesn't change while on TagsPage. If it does changed (we just switched from a different page), `oncreate` will run. So `onupdate` shouldn't be needed.

js/src/forum/components/TagsPage.js
askvortsov1: I believe that's just a difference between `m.route.Link` and `m.route` as config on links. In the old version, if you had a link inside a link, both would execute, so stopPropogation was necessary. Now, due to how m.route.Link works, that stopPropogation is built in I believe (or it otherwise prevents links inside links from opening twice).

js/src/forum/addTagList.js
askvortsov1: That being said, it DOES work! :partying_face: :tada:

js/src/forum/components/TagDiscussionModal.js
askvortsov1: It's a workaround from before when `app.modal.close` was broken. That being said, I kind of like how it is now, directly calling `this.hide()` instead of calling it through the global `ModalState`

js/src/forum/components/TagsPage.js
clarkwinkelmann: I'm not sure I've seen this before but @askvortsov1 explanation makes sense. It must be due to the nested `<a>` tags, and might no longer be needed.

franzliedke (approved):

js/src/forum/components/TagDiscussionModal.js
franzliedke: But doing so does not mark the modal as closed in the state, right?

askvortsov1: Well, if that's an issue here, that also affects closing modals by clicking (x), as that also calls this.hide().

askvortsov1: Perhaps we should build in something to clear the modal without a redraw in ModalManager's animateHide?

franzliedke: Hmm, fair point. I guess we can live with it here, but both cases should be fixed at some point, as they would both mean state and view getting out of sync?

askvortsov1: Agreed, I'll make an issue. It doesn't seem to affect the UI, but it's still something that would be preferable to fix.

PR #94, opened 2020-09-22
Remove call to undefined method handleErrors

w-4:
Fixes admin erorr: `Uncaught TypeError: e.handleErrors is not a function`

Modal.handleErrors has been removed 5 years ago and became Modal.onerror (commit id: `26a821e3e28df83128d7d883a98b6cea129375a8`).

### DIFF
--- a/js/src/admin/components/EditTagModal.js
+++ b/js/src/admin/components/EditTagModal.js
@@ -123,6 +123,8 @@ export default class EditTagModal extends Modal {
 
     this.loading = true;
 
+    // Errors aren't passed to the modal onerror handler here. 
+    // This is done for better error visibility on smaller screen heights.
     this.tag.save(this.submitData()).then(
       () => this.hide(),
       () => this.loading = false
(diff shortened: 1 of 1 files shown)


askvortsov1 (commented):
Thank you!!! One question: when it does error, will that error still bubble up to the onerror handler?

w-4:
@askvortsov1 not to the modal onerror. It shows the error message on the bottom left of the screen. Maybe that's actually better, since the tag modal doesn't fit the page on mobile, so you need to scroll down to save changes. If the error renders inside the modal after submitting, there is no instant visual feedback, you'd need to scroll up to see it. 
But really I don't have a preference, either is fine for me.

askvortsov1:
Considering the height of the modal, I am fine with having this modal behave different from others. That being said, could you add a comment explaining that errors won't be passed to onerror for better UX?

Thanks for yet another great PR!

askvortsov1 (approved):
Looks good, thank you very much!

clarkwinkelmann (approved):
I would have preferred to see an issue with steps to reproduce first, but I do understand what the issue is and I just confirmed the issue exists locally on beta 13.

I have not tested this PR but the change looks good.

PR #95, opened 2020-10-03
Fix Editing Discussion Tags Permission

SychO9:
**Fixes flarum/core#2355**


askvortsov1 (approved):
Tested locally, and it works. Also intuitively makes sense (this is the same logic as is used to check whether the edit tags control option should even be displayed). Thanks for the quick PR!

clarkwinkelmann (approved):
Looking good code-wise, and I trust @askvortsov1 with the local tests.

PR #96, opened 2020-10-05
Use Link component instead of route patch for tagLabel links

w-4:
Use Link component instead of route patch
see: https://github.com/flarum/tags/commit/56c4c3e16624538cd0c15a0b224fb440ef82b658

- [x] tested locally


datitisev (approved):

askvortsov1 (approved):
Great catch, thanks!

PR #97, opened 2020-10-06
Use AssertPermissionTrait instead of $actor->can in backported fix

askvortsov1:
$actor->can was introduced in beta 14, and is not yet available.


luceos (approved):

clarkwinkelmann (approved):
Tested locally on beta 13 :+1:

PR #101, opened 2020-11-12
AdminUX Update

KyrneDev:
AdminUX update ref: https://github.com/flarum/core/pull/2409

Screen Shot 2020-11-12 at 10 22 14 AM


askvortsov1 (changes requested):
Please revert unnecessary stylistic changes. We need to configure prettier for bundled extensions, but till then, let's avoid that.

PR #107, opened 2020-12-25
Fix starting discussion without tags

SychO9:
**Fixes flarum/core#2510**
**Fixes flarum/core#2511**

Cleans up the code to properly handle the logic for tags when starting a discussion or editing one. I'm not sure if it would be better to separate the code into a creation method and an editing method, that might duplicate some code, but be more readable.

<details>
<summary>Logic Details</summary>
<p>

## Current code logic
```
- when a tags relationship is provided:
  - when editing a discussion:
    - check for `tag` ability on the discussion
  - check `startDiscussion` permission on each tag
  - validate tag count
  - when editing a discussion:
    - check for `addTagToDiscussion` permission on each _new_ tag
    - raise `DiscussionWasTagged` event
  - sync tags relationship
- when a tags relationship is not provided and a new discussion is being started and actor doesn't have permission to `startDiscussion`
^^^ Only works if startDiscussion permission is false and a tags relationship has not been provided at all (if you provide an empty tags relationship it'll fail as well)
```

## How it should be
**Start Discussion**
```
- check ability to `startDiscussion` in each tag
- validate tag count
- sync tags relationship
```
**Edit Discussion**
```
- when a tags relationship is provided:
  - check ability to `tag` the discussion
  - check ability to `addToDiscussion` for each _new_ tag
  - validate tag count
  - raise `DiscussionWasTagged` event
  - sync tags relationship
```

## New Code Logic
```
- when a tags relationship is provided:
  - make an array of new tag models
- when editing and a tags relationship is provided:
  - check ability to `tag` discussion
  - check ability to `addToDiscussion` for each _new_ tag
  - raise `DiscussionWasTagged` event
- when starting or a tags relationship is provided:
  - when starting a discussion:
    - check ability to `startDiscussion` on each tag
  - validate tag count
  - sync tags relationship
```

</p>

### DIFF
--- a/src/Listener/SaveTagsToDatabase.php
+++ b/src/Listener/SaveTagsToDatabase.php
@@ -57,25 +57,46 @@ class SaveTagsToDatabase
         $discussion = $event->discussion;
         $actor = $event->actor;
 
-        // TODO: clean up, prevent discussion from being created without tags
-        if (isset($event->data['relationships']['tags']['data'])) {
-            if ($discussion->exists) {
-                $actor->assertCan('tag', $discussion);
-            }
+        $newTagIds = [];
+        $newTags = [];
 
+        $primaryCount = 0;
+        $secondaryCount = 0;
+
+        if (isset($event->data['relationships']['tags']['data'])) {
             $linkage = (array) $event->data['relationships']['tags']['data'];
 
-            $newTagIds = [];
             foreach ($linkage as $link) {
                 $newTagIds[] = (int) $link['id'];
             }
 
             $newTags = Tag::whereIn('id', $newTagIds)->get();
-            $primaryCount = 0;
-            $secondaryCount = 0;
+        }
+
+        if ($discussion->exists && isset($event->data['relationships']['tags']['data'])) {
+            $actor->assertCan('tag', $discussion);
+
+            $oldTags = $discussion->tags()->get();
+            $oldTagIds = $oldTags->pluck('id')->all();
+
+            if ($oldTagIds == $newTagIds) {
+                return;
+            }
 
             foreach ($newTags as $tag) {
-                if ($actor->cannot('startDiscussion', $tag)) {
+                if (! in_array($tag->id, $oldTagIds) && $actor->cannot('addToDiscussion', $tag)) {
+                    throw new PermissionDeniedException;
+                }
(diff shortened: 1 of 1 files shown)


src/Listener/SaveTagsToDatabase.php
```diff
@@ -89,30 +110,9 @@ public function handle(Saving $event)
             $this->validateTagCount('primary', $primaryCount);
```
askvortsov1, line 110: Either as part of this (or preferably as a separate PR that would be merged beforehand), I think we should implement https://github.com/flarum/core/issues/2138 before this is merged. That way, byobu can still create discussions without tags through a policy. All we'd need to do is wrap the 2 validate lines:

```
if (!$actor->can('bypassTagCounts`)) {
    $this->validateTagCount('primary', $primaryCount);
    $this->validateTagCount('secondary', $secondaryCount);
}
```

src/Listener/SaveTagsToDatabase.php
```diff
@@ -89,30 +110,9 @@ public function handle(Saving $event)
...
-
-                $discussion->raise(
-                    new DiscussionWasTagged($discussion, $actor, $oldTags->all())
-                );
-            }
-
             $discussion->afterSave(function ($discussion) use ($newTagIds) {
                 $discussion->tags()->sync($newTagIds);
             });
-        } elseif (! $discussion->exists && ! $actor->hasPermission('startDiscussion')) {
```
askvortsov1, line 114: Would https://github.com/flarum/tags/pull/108 be replacing this bit?

SychO9: I separated it because it fixed a different issue, and removing the bit here didn't make a difference.
Looking at the code, it does indeed make more sense now to include that here instead.

src/Listener/SaveTagsToDatabase.php
SychO9: Will do,

so byobu currently just makes use of this bug to have tagless discussions created ?

askvortsov1: Yep

askvortsov1 (approved):

tankerkiller125 (approved):

PR #108, opened 2020-12-25
Fix starting a discussion without tags, without global startDiscussion perm

SychO9:
**Fixes flarum/core#2511**

It's a special case scenario, where the user in blue role has the **ability** to `startDiscussion` (because he has enough primary and secondary tags discussion starting permissions). But does not have the **permission** to globally `startDiscussion`, so we have to explicitly check that when no tags are given and the user doesn't have the permission, we deny access.

This isn't something we can move to the policy because the policy does not know how many tags the user has provided.

### DIFF
--- a/src/Listener/SaveTagsToDatabase.php
+++ b/src/Listener/SaveTagsToDatabase.php
@@ -114,6 +114,10 @@ class SaveTagsToDatabase
                 $discussion->tags()->sync($newTagIds);
             });
         }
+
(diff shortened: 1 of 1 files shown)


askvortsov1:
Is this intended to be merged after #107? If so, why not include it there, since it's replacing something that's removed there?

SychO9:
https://github.com/flarum/tags/pull/107#discussion_r552068079

PR #111, opened 2021-01-08
Add permission to bypass tag count requirements

SychO9:
**Fixes flarum/core#2138**

Before I PR the english repo, would we rather have the permission in "Create" category named something like "Start discussion without tag count requirements" ? or in "Moderate" category with something like "Bypass minimum number of tags required" ?

You'll notice that I didn't change the behavior of the TagDiscussionModal showing up when the min requirements are not met even if you have permission. They can still submit, that way it is more explicit that they are bypassing the required amount of tags.

* [x] flarum/lang-english#182

### DIFF
--- a/extend.php
+++ b/extend.php
@@ -47,7 +47,10 @@ return [
         ->belongsToMany('tags', Tag::class, 'discussion_tag'),
 
     (new Extend\ApiSerializer(ForumSerializer::class))
-        ->hasMany('tags', TagSerializer::class),
+        ->hasMany('tags', TagSerializer::class)
+        ->attribute('canBypassTagCounts', function (ForumSerializer $serializer) {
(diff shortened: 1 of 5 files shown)


askvortsov1 (commented):
I like it under start, but labelled as "bypass tag requirements". Should https://github.com/flarum/core/issues/2473 be addressed here too? We might see more of that with this PR since some people will be allowed to start discussions even if there are no tags at all.

SychO9:
Good catch, although I would prefer to fix that in a separate pull request since it doesn't necessarily have to go here and it would be useful to have things more organized for future convenience.

As a side note, I wouldn't try to test this pull request yet, at least until #106 is merged as it might get confusing.

askvortsov1 (approved):

tankerkiller125 (approved):

PR #113, opened 2021-01-24
Add missing Button-label span from ButtonLink

davwheat:
Every side nav item created by flarum/core (e.g. All Discussions, etc.) includes this span, so tags should too.


askvortsov1 (approved):
Tested locally, doesn't seem to break anything

SychO9 (approved):
Thank you for the PR!

PR #114, opened 2021-02-01
Add "TagLabel--child" class to aid with custom theming

davwheat:
I've run into issues with theming tags in discussion lists and other areas as there's nothing in the DOM that can be used to differentiate between primary/secondary/child tags.

Adding this custom data attribute would be super helpful for theming, especially with extensions like night mode where keeping the tags the same colour between light/dark could make them unreadable.

----

**PS:** The repo seems to be missing the Prettier stuff that other core repos have. Should I PR it all in?

### DIFF
--- a/js/src/common/helpers/tagLabel.js
+++ b/js/src/common/helpers/tagLabel.js
@@ -21,7 +21,9 @@ export default function tagLabel(tag, attrs = {}) {
       attrs.href = app.route('tag', {tags: tag.slug()});
     }
 
-    attrs["data-is-child"] = tag.data.attributes.isChild ? "true" : "false";
+    if (tag.data.attributes.isChild()) {
+      attrs['data-is-child'] = 'true'
+    }
   } else {
     attrs.className += ' untagged';
   }
--- a/js/src/common/helpers/tagLabel.js
+++ b/js/src/common/helpers/tagLabel.js
@@ -22,7 +22,7 @@ export default function tagLabel(tag, attrs = {}) {
     }
 
     if (tag.isChild()) {
-      attrs.className += ' TagLabel--child'
+      attrs.className += ' TagLabel--child';
     }
   } else {
     attrs.className += ' untagged';
--- a/js/src/common/helpers/tagLabel.js
+++ b/js/src/common/helpers/tagLabel.js
@@ -1,38 +1,37 @@
-import extract from "flarum/utils/extract";
-import Link from "flarum/components/Link";
-import tagIcon from "./tagIcon";
+import extract from 'flarum/utils/extract';
+import Link from 'flarum/components/Link';
+import tagIcon from './tagIcon';
 
 export default function tagLabel(tag, attrs = {}) {
   attrs.style = attrs.style || {};
-  attrs.className = "TagLabel " + (attrs.className || "");
+  attrs.className = 'TagLabel ' + (attrs.className || '');
 
-  const link = extract(attrs, "link");
-  const tagText = tag
-    ? tag.name()
-    : app.translator.trans("flarum-tags.lib.deleted_tag_text");
+  const link = extract(attrs, 'link');
+  const tagText = tag ? tag.name() : app.translator.trans('flarum-tags.lib.deleted_tag_text');
 
   if (tag) {
     const color = tag.color();
     if (color) {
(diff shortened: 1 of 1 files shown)


js/src/common/helpers/tagLabel.js
```diff
@@ -20,6 +20,9 @@ export default function tagLabel(tag, attrs = {}) {
       attrs.title = tag.description() || '';
       attrs.href = app.route('tag', {tags: tag.slug()});
     }
+
+    attrs["data-is-child"] = tag.data.attributes.isChild ? "true" : "false";
+    console.log(tag);
```
SychO9, line 25: you forgot the debugging code :D

SychO9 (changes requested):

js/src/common/helpers/tagLabel.js
davwheat: Oh of course I did! 😂

datitisev:
Instead of having it equal `true` or `false`, I'd rather have it present or not present. Thoughts?

Also - make sure you use single quotes for strings.

davwheat:
> Instead of having it equal `true` or `false`, I'd rather have it present or not present. Thoughts?

Sounds good. I'll do `data-is-child=true` and not present, as that seems to match where this sort of stuff is done elsewhere in core.

> Also - make sure you use single quotes for strings.

Oops. I'm gonna make a PR to add `.prettierrc` and the format script because I keep forgetting.

SychO9:
why not use a class instead ? `TagLabel--child`

datitisev:
Actually, what @SychO9 makes more sense, my bad. We use classes for mostly everything, stuff like if the Post was made by the current user, etc... so `TagLabel--child` does make more sense.

davwheat:
Ah, that's true. I've always been a fan of data attributes over classes for identifying features like this (generally because they're tstable with `^=` or `*=`, but since classes are used for everything, that's probably a better idea.

js/src/common/helpers/tagLabel.js
```diff
@@ -20,6 +20,10 @@ export default function tagLabel(tag, attrs = {}) {
       attrs.title = tag.description() || '';
       attrs.href = app.route('tag', {tags: tag.slug()});
     }
+
+    if (tag.data.attributes.isChild()) {
+      attrs.className += 'TagLabel--child'
```
datitisev, line 25: There needs to be a space before the `TagLabel--child` like with `untagged`

datitisev (changes requested):

js/src/common/helpers/tagLabel.js
davwheat: I'm making the dumbest mistakes today, my bad. Fixed now.

js/src/common/helpers/tagLabel.js
```diff
@@ -20,6 +20,10 @@ export default function tagLabel(tag, attrs = {}) {
       attrs.title = tag.description() || '';
       attrs.href = app.route('tag', {tags: tag.slug()});
     }
+
+    if (tag.data.attributes.isChild()) {
```
askvortsov1, line 24: Why not `tag.isChild()`?

askvortsov1 (changes requested):

askvortsov1 (approved):

js/src/common/helpers/tagLabel.js
```diff
@@ -20,6 +20,10 @@ export default function tagLabel(tag, attrs = {}) {
       attrs.title = tag.description() || '';
       attrs.href = app.route('tag', {tags: tag.slug()});
     }
+
+    if (tag.isChild()) {
+      attrs.className += ' TagLabel--child'
```
SychO9, line 25: nitpick: you forgot the semicolon at the end :p `;`

askvortsov1: Huh, I thought the JS lint GH action would have caught that...

SychO9: I don't think that's setup here, there's just the build workflow

davwheat: That's a good spot -- none of the styling stuff has been set up here except StyleCI. Only the flarum build bot is on here.

SychO9 (approved):

PR #115, opened 2021-02-01
Add Prettier formatting support

davwheat:
The repo was missing Prettier support and config, so this PR adds it and uses it.

It's set up exactly how it's done in flarum/core.

### DIFF
--- a/js/.prettierrc
+++ b/js/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "printWidth": 150,
+  "singleQuote": true,
+  "tabWidth": 2,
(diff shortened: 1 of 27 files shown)


askvortsov1:
Thanks for the PR! I think we might hold off on this one for a bit, as we're considering changing core to remove the husky pre commit hook, so I don't think we should add it for bundled extensions until that decision is made. We do plan to eventually add this in though.

askvortsov1:
I'm going to close this, as it'll be simpler to do in one go closer to the release

PR #116, opened 2021-02-04
Max-height for tags settings

KyrneDev:
The form (3rd column) of the tags page can get super long with lots of tags as its height is related to the tags lists. This limits its height to a reasonable one. 

Core: https://github.com/flarum/core/pull/2593


askvortsov1 (approved):

askvortsov1:
Tested locally, works well.

tankerkiller125 (approved):

PR #119, opened 2021-02-17
Floating Discussion Title

KyrneDev:
Hides non-primary tags in the floating hero to prevent a tags overload! Also as a bonus adds a class for theming. Editing tags still works :)

ref: https://github.com/flarum/core/pull/2606

### DIFF
--- a/less/admin/TagsPage.less
+++ b/less/admin/TagsPage.less
@@ -96,6 +96,7 @@ li:not(.sortable-dragging)>.TagListItem-info:hover>.Button {
 
   .Form {
     min-width: 300px;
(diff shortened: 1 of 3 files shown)


less/admin/TagsPage.less
```diff
@@ -96,6 +96,7 @@ li:not(.sortable-dragging)>.TagListItem-info:hover>.Button {
 
   .Form {
     min-width: 300px;
+    max-height: 500px;
```
askvortsov1, line 99: This should be removed, I think it was already merged in a separate PR?

PR #120, opened 2021-02-24
[Performance] Eagerload lastPostedDiscussion needed relationships

SychO9:
**References flarum/core#2177**

Explained in https://github.com/flarum/core/issues/2177#issuecomment-785256237.


askvortsov1 (approved):

luceos:
Understand that we can only eager load the state relation if `stateFor` was called ahead of that. I'm glad it was, so approving.

luceos (approved):

PR #121, opened 2021-03-01
made compatible with illuminate 8 and beta 16

luceos:
depends on https://github.com/flarum/core/pull/2576


askvortsov1 (approved):
Ran into this locally too when testing, this change fixed it.

PR #127, opened 2021-04-24
Permission Tag Scoping Field

askvortsov1:
Part of https://github.com/flarum/core/issues/2202

This allows tag scoping permissions that don't start with `discussion`, or on the contrary, removing tag scoping from permissions that DO start with `discussion`.

### DIFF
--- a/js/src/admin/addTagsPermissionScope.js
+++ b/js/src/admin/addTagsPermissionScope.js
@@ -30,6 +30,8 @@ export default function() {
         label: tagLabel(tag),
         onremove: () => tag.save({isRestricted: false}),
         render: item => {
+          if (item.tagScoped !== undefined) return item.tagScoped;
+          
           if (item.permission === 'viewDiscussions'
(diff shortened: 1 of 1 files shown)


clarkwinkelmann:
How does this work? Wouldn't this return just `true` instead of an expected instance of `PermissionDropdown` ?

js/src/admin/addTagsPermissionScope.js
```diff
@@ -29,10 +29,11 @@ export default function() {
       .forEach(tag => items.add('tag' + tag.id(), {
         label: tagLabel(tag),
         onremove: () => tag.save({isRestricted: false}),
-        render: item => {
-          if (item.permission === 'viewDiscussions'
+        render: item => {
```
SychO9, line 32:
```suggestion
        render: item => {
```

SychO9 (approved):
Tested locally, code looks good.

PR #128, opened 2021-05-05
Refactor queryIdsWhereCan

askvortsov1:
- Rename to queryIdsWhereHasPermission, since that's more accurate
- Make the base query an optional 3rd argument. This feels more intuitive
- Add a `can` function in AbstractPolicy, so extensions that add tag scopable permissions don't need to define their own policies (e.g. with approval)

This makes https://github.com/flarum/approval/pull/27 obsolete.

### DIFF
--- a/src/Access/GlobalPolicy.php
+++ b/src/Access/GlobalPolicy.php
@@ -37,8 +37,8 @@ class GlobalPolicy extends AbstractPolicy
             if ($actor->hasPermission($ability) && $actor->hasPermission('bypassTagCounts')) {
                 return $this->allow();
             }
-            $enoughPrimary = Tag::queryIdsWhereCan(Tag::query()->getQuery(), $actor, $ability, true, false)->count() >= $this->settings->get('flarum-tags.min_primary_tags');
-            $enoughSecondary = Tag::queryIdsWhereCan(Tag::query()->getQuery(), $actor, $ability, false, true)->count() >= $this->settings->get('flarum-tags.min_secondary_tags');
+            $enoughPrimary = Tag::queryIdsWhereHasPermission($actor, $ability, null, true, false)->count() >= $this->settings->get('flarum-tags.min_primary_tags');
+            $enoughSecondary = Tag::queryIdsWhereHasPermission($actor, $ability, null, false, true)->count() >= $this->settings->get('flarum-tags.min_secondary_tags');
 
             if ($enoughPrimary && $enoughSecondary) {
                 return $this->allow();
--- a/src/Access/ScopeDiscussionVisibilityForAbility.php
+++ b/src/Access/ScopeDiscussionVisibilityForAbility.php
@@ -33,7 +33,7 @@ class ScopeDiscussionVisibilityForAbility
             return $query->select('discussion_id')
                 ->from('discussion_tag')
                 ->whereIn('tag_id', function ($query) use ($actor, $ability) {
-                    Tag::queryIdsWhereCan($query->from('tags'), $actor, 'discussion.'.$ability);
+                    Tag::queryIdsWhereHasPermission($actor, 'discussion.'.$ability, $query->from('tags'));
                 });
         });
     }
--- a/src/Access/TagPolicy.php
+++ b/src/Access/TagPolicy.php
@@ -15,25 +15,10 @@ use Flarum\User\User;
 
 class TagPolicy extends AbstractPolicy
 {
-    /**
-     * @param User $actor
-     * @param Tag $tag
-     * @return bool|null
-     */
-    public function startDiscussion(User $actor, Tag $tag)
+    public function can(User $actor, string $ability, Tag $tag)
     {
         if ($tag->is_restricted) {
-            return $actor->hasPermission('tag'.$tag->id.'.startDiscussion') ? $this->allow() : $this->deny();
+            return Tag::queryIdsWhereHasPermission($actor, $ability)->where('id', $tag->id)->count() !== 0;
         }
-    }
-
-    /**
-     * @param User $actor
-     * @param Tag $tag
-     * @return bool|null
-     */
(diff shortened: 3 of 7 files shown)


src/Access/TagPolicy.php
```diff
@@ -15,25 +15,15 @@
...
-     * @param User $actor
-     * @param Tag $tag
-     * @return bool|null
-     */
-    public function startDiscussion(User $actor, Tag $tag)
+    public function can(User $actor, string $ability, Tag $tag)
     {
         if ($tag->is_restricted) {
-            return $actor->hasPermission('tag'.$tag->id.'.startDiscussion') ? $this->allow() : $this->deny();
+            return Tag::queryIdsWhereHasPermission($actor, $ability)->where('id', $tag->id)->count() !== 0;
```
SychO9, line 21: I understand why we're switching to `can` but not why we're using `queryIdsWhereHasPermission` instead of `hasPermission` since we're then doing a `where('id', $tag->id)` ? or is it because we also need to check that the parent isn't restricted ?

my main concern is that this is a query for every tag.

askvortsov1: > or is it because we also need to check that the parent isn't restricted ?

Yeah. Now that you raise that issue, we should probably keep it as `hasPermission`.

clarkwinkelmann:
The changes in `TagPolicy` with the tag parent thing are another change not described in your first comment, no? There's a change of behavior there I suppose?

Wouldn't the `queryIdsWhereCan`/`queryIdsWhereHasPermission` method be better as a query scope? I suppose a visibility scope isn't possible since it needs additional parameters, but a simple Eloquent model scope could also be an option?

askvortsov1:
Updated, it's a lot more versatile now.

src/Access/GlobalPolicy.php
```diff
@@ -37,8 +37,12 @@ public function can(User $actor, string $ability)
...
                 return $this->allow();
             }
-            $enoughPrimary = Tag::queryIdsWhereCan(Tag::query()->getQuery(), $actor, $ability, true, false)->count() >= $this->settings->get('flarum-tags.min_primary_tags');
-            $enoughSecondary = Tag::queryIdsWhereCan(Tag::query()->getQuery(), $actor, $ability, false, true)->count() >= $this->settings->get('flarum-tags.min_secondary_tags');
+            $enoughPrimary = Tag::whereHasPermission($actor, $ability)
+                ->where('tags.position', '!=', null)
+                ->count() >= $this->settings->get('flarum-tags.min_primary_tags');
+            $enoughSecondary = Tag::whereHasPermission($actor, $ability)
+                ->where('tags.position', '=', null)
+                ->count() >= $this->settings->get('flarum-tags.min_secondary_tags');
```
SychO9, line 45: damn, one thing I hadn't noticed before, is that this also creates multiple queries now.. I think we can cache the result and reuse that ?

src/Access/TagPolicy.php
```diff
@@ -15,25 +15,21 @@
...
-     * @param User $actor
-     * @param Tag $tag
-     * @return bool|null
-     */
-    public function startDiscussion(User $actor, Tag $tag)
+    public function can(User $actor, string $ability, Tag $tag)
     {
+        if ($tag->parent_id !== null && ! $actor->can($ability, $tag->parent)) {
+            return $this->deny();
+        }
```
SychO9, line 22: now that we need the parent here *( I'm still not caught up with why, I suppose it has to do with the approval extension kinda, that's what I'm gonna check next )* can we eager load the `tags.parent` relationship in `ListDiscussionsController`  by adding it in `extend.php` and force load it in `ListTagsController` ?

SychO9 (changes requested):
I love that we have tests to make sure all the different scenarios are properly working, I see there's quiet a number of tests, do they cover discussion visibility ?

src/Access/ScopeDiscussionVisibilityForAbility.php
```diff
@@ -29,11 +29,11 @@ public function __invoke(User $actor, Builder $query, $ability)
         // If a discussion requires a certain permission in order for it to be
         // visible, then we can check if the user has been granted that
         // permission for any of the discussion's tags.
-        $query->whereIn('discussions.id', function ($query) use ($actor, $ability) {
+        $query->whereIn('discussions.id', function ($query) use ($actor) {
```
clarkwinkelmann, line 32: Why is the `$ability` value ignored now? Doesn't it defeat the purpose of this scope?

src/Tag.php
```diff
@@ -240,30 +235,21 @@ public static function queryIdsWhereHasPermission(User $user, string $currPermis
...
-                if (! $includePrimary) {
-                    $query->where('s_tags.position', '=', null);
-                }
-                if (! $includeSecondary) {
-                    $query->where('s_tags.position', '!=', null);
-                }
             });
 
-        return $validTags->select('s_tags.id');
+        return $validTags;
```
clarkwinkelmann, line 253: The `return` could be moved to line 238 and the `$validTags` variable skipped.

src/Access/GlobalPolicy.php
askvortsov1: Not sure that we can do this: we'd have to cache per user. Also, why would it be multiple queries?

SychO9: Right now we make 4 queries, 2 for each (viewDiscussion, startDiscussion) + other repeated onces whenever the global ability is checked again other than the ForumSerializer.

I've went ahead and cached it like we do in approval's TagPolicy, I think it makes sense, queries went down from 46 to 32 locally, not sure where else they were called.

clarkwinkelmann (approved):
That seems ok for me.

I haven't tested locally. I see some odd changes to the tests (response codes change?) but I trust they make sense :innocent:

PR #130, opened 2021-05-11
Only check for bypassTagCounts permission for startDiscussion ability

SychO9:
This doesn't fix anything, but we should explicitly only check the `bypassTagCounts` permission for the ability it is meant for (`startDiscussion`), just to be safe in the future.

### DIFF
--- a/tests/integration/authorization/GlobalPolicyTest.php
+++ b/tests/integration/authorization/GlobalPolicyTest.php
@@ -152,7 +152,6 @@ class GlobalPolicyTest extends TestCase
     public function can_start_discussion_globally_if_sufficient_tags_and_allows_regardless_of_start_discussion_and_bypass()
     {
         $this->database()->table('group_permission')->where('permission', 'bypassTagCounts')->delete();
-        $this->database()->table('group_permission')->where('permission', 'startDiscussion')->delete();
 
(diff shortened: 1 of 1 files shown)


tests/integration/authorization/GlobalPolicyTest.php
```diff
@@ -113,4 +113,49 @@ public function can_start_discussion_globally_if_allowed_in_secondary_tag_and_mi
 
         $this->assertTrue(User::find(2)->can('startDiscussion'));
     }
+
+    /**
+     * @test
+     */
+    public function cant_start_discussion_globally_if_permission_not_granted_with_granted_bypass_permission()
```
askvortsov1, line 120: Could we rename these to "if_allowed_in_sufficient_tags" (and not allowed) instead of permission granted?

askvortsov1 (changes requested):

SychO9:
Tests are failing, but I only renamed the methods, so I suspect it has to do with the fact that we just renamed the two global permissions in core, and we always use latest dev-master when testing bundled extensions.

Which is something we'll need to look into organizing possibly better after 2.0 (don't have any ideas atm)

askvortsov1 (approved):

PR #132, opened 2021-05-17
Check that `included` array exists

SychO9:
Secondary tag document has no array of `included` tags.

### DIFF
--- a/src/Content/Tag.php
+++ b/src/Content/Tag.php
@@ -82,7 +82,7 @@ class Tag
         $tagsDocument = $this->getTagsDocument($request, $slug);
 
         $apiDocument->included[] = $tagsDocument->data;
-        $includedTags = $tagsDocument->included;
+        $includedTags = $tagsDocument->included ?? [];
         foreach ((array) $includedTags as $includedTag) {
             $apiDocument->included[] = $includedTag;
         }


src/Content/Tag.php
```diff
@@ -82,7 +82,7 @@ public function __invoke(Document $document, Request $request)
         $tagsDocument = $this->getTagsDocument($request, $slug);
 
         $apiDocument->included[] = $tagsDocument->data;
-        $includedTags = $tagsDocument->included;
+        $includedTags = $tagsDocument->included ?? [];
```
askvortsov1, line 85: Isn't the issue that the `included` property might not be defined? Would the null coalesce operator handle this? I know PHP8 introduced nullsafe for pretty much this reason.

SychO9: the null coalesce operator in php is basically just a shorthand for `isset($var) ? $var : ..` and `isset()` can handle this.

SychO9: also yes, in secondary tags, it is not defined

askvortsov1: Ah cool then! I love it when syntax does things that wouldn't be expected from it in a sane programming language. Thanks PHP!

askvortsov1 (approved):

PR #133, opened 2021-06-01
Load all tags in permission grid

SychO9:
**Fixes flarum/core#2891**

Loads all tags before rendering the grid since we now only load a subset with the forum serializer.


luceos (approved):
This now works, great job @SychO9

PR #134, opened 2021-06-07
Fix wrong subtags listing in sidebar

SychO9:
**Fixes flarum/core#2895**

The missing parent relationship in the tags page lead to the filtering of the active tag's subtags to return a wrong `true` value, the parent relationship is necessary.


the-turk:
sadly, this PR didn't solve the listing issue completely

1. Go to Tags page
2. Refresh the page
3. Head into any primary tag
4. See the mess

QUADEMU:
@the-turk
i confirm by that result after select META primary tag
image

and for go more deeply without open an new issue, the issue can be repeat by select any primary tag in the two pos [0-1], that forward you in all discussions with an messy tags order list.
regards

PR #135, opened 2021-06-10
resolves the issue with secondaries loading sporadically

luceos:
When starting the forum from anything but the index and you go to the index and then any other tag, all secondaries in the js store are shown. This is caused by a missing parent include when hitting the index.

This also seems to fix flarum/core#2907


davwheat (approved):

the-turk:
see https://github.com/flarum/tags/pull/134#issuecomment-861665957

QUADEMU:
> see [#134 (comment)](https://github.com/flarum/tags/pull/134#issuecomment-861665957)

image

PR #138, opened 2021-06-29
Make it easier for other extensions to extend TagsPage

andreasjacobsen93:
This PR proposes a change to separate logic for retrieving tags to its own function to allow other extensions to override that function and manipulate tags on TagsPage.

### DIFF
--- a/js/src/forum/components/TagsPage.js
+++ b/js/src/forum/components/TagsPage.js
@@ -17,22 +17,29 @@ export default class TagsPage extends Page {
 
     this.tags = [];
 
-    const preloaded = app.preloadedApiDocument();
-
-    if (preloaded) {
-      this.tags = sortTags(preloaded.filter(tag => !tag.isChild()));
-      return;
-    }
-
     this.loading = true;
 
     app.tagList.load(['children', 'lastPostedDiscussion', 'parent']).then(() => {
-      this.tags = sortTags(app.store.all('tags').filter(tag => !tag.isChild()));
+      this.tags = this.items();
 
       this.loading = false;
 
       m.redraw();
     });
+  }
+  
+  items() {
+    if (this.tags.length > 0) {
+      return this.tags;
+    }
+    
+    const preloaded = app.preloadedApiDocument();
+
+    if (preloaded) {
+      return sortTags(preloaded.filter(tag => !tag.isChild()));
+    }
+    
+    return sortTags(app.store.all('tags').filter(tag => !tag.isChild()));
   }
 
   view() {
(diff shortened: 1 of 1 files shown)


SychO9:
Hello @andreasjacobsen93, sorry for the wait, thank you for the PR, we'll review this soon!

js/src/forum/components/TagsPage.js
```diff
@@ -17,31 +17,38 @@ export default class TagsPage extends Page {
...
 
   view() {
     if (this.loading) {
       return <LoadingIndicator />;
     }
 
-    const pinned = this.tags.filter(tag => tag.position() !== null);
-    const cloud = this.tags.filter(tag => tag.position() === null);
+    const pinned = this.items().filter(tag => tag.position() !== null);
+    const cloud = this.items().filter(tag => tag.position() === null);
```
luceos, line 51: I would prefer assigning the result of `this.items()` to a `const items` or use `this.items` as that property should also be available by then and then dissecting that into `pinned` and `cloud`. Extensions might extend the `items` method and cause increasing overhead; hitting that method then twice in a row while the result is exactly the same doesn't make much sense.

luceos (changes requested):

js/src/forum/components/TagsPage.js
andreasjacobsen93: Great suggestion, thank you :)

js/src/forum/components/TagsPage.js
```diff
@@ -17,31 +17,39 @@ export default class TagsPage extends Page {
...
+  }
 
   view() {
     if (this.loading) {
       return <LoadingIndicator />;
     }
 
-    const pinned = this.tags.filter(tag => tag.position() !== null);
-    const cloud = this.tags.filter(tag => tag.position() === null);
+    const items = this.items()
```
davwheat, line 50: Could we not use the value of `this.tags` which is set in `oninit`? This saves executing `this.items()` twice.

js/src/forum/components/TagsPage.js
```diff
@@ -17,31 +17,39 @@ export default class TagsPage extends Page {
...
-      this.tags = sortTags(app.store.all('tags').filter(tag => !tag.isChild()));
+      this.tags = this.items();
 
       this.loading = false;
 
       m.redraw();
     });
   }
+  
+  items() {
```
askvortsov1, line 31: I would prefer to avoid using `items` for methods that don't use ItemLists

js/src/forum/components/TagsPage.js
```diff
@@ -17,31 +17,39 @@ export default class TagsPage extends Page {
...
+    if (this.tags.length > 0) {
+      return this.tags;
+    }
+    
+    const preloaded = app.preloadedApiDocument();
+
+    if (preloaded) {
```
askvortsov1, line 38: Could we use a ternary assignment, so we don't repeat the filtering and sortTags logic?

askvortsov1 (changes requested):
Thanks for the PR! I'm not sure I like this approach, because it obfuscates data flow and sourcing a bit: there's nothing enforcing that items shouldn't be called until tagList.load is called.

Maybe we could move that entire block from oninit to a `loadTags` method ala `this.tags = await this.loadTags` (or the promise equivalent)?

Could you share a bit more about what transformations you want to perform while loading tags too?

PR #139, opened 2021-08-06
[1.x] Custom Colorising with CSS Custom Properties

SychO9:
**Goes with flarum/core#3001**

### DIFF
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -69,7 +69,8 @@ export default function() {
       const canStartDiscussion = tag.canStartDiscussion() || !app.session.user;
 
       if (color) {
-        items.get('newDiscussion').attrs.style = {backgroundColor: color};
+        items.get('newDiscussion').attrs.className += ' Button--tagColored';
+        items.get('newDiscussion').attrs.style = { '--color': color };
       }
 
       items.get('newDiscussion').attrs.disabled = !canStartDiscussion;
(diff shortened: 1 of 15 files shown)


js/src/forum/addTagFilter.js
```diff
@@ -69,7 +69,8 @@ export default function() {
       const canStartDiscussion = tag.canStartDiscussion() || !app.session.user;
 
       if (color) {
-        items.get('newDiscussion').attrs.style = {backgroundColor: color};
+        items.get('newDiscussion').attrs.className += ' Button--tagColored';
```
davwheat, line 72: I'd be more inclined to use `classList` here, personally. It makes no functional different, but means that we won't end up with possible extra whitespace in the class attribute.

IMO the extra processing time is very much negligible for the benefit of consistency.

js/src/forum/components/TagLinkButton.js
```diff
@@ -17,9 +17,9 @@ export default class TagLinkButton extends LinkButton {
 
     return (
       <Link className={className} href={this.attrs.route}
-        style={active && tag ? {color: tag.color()} : ''}
+        style={tag ? { '--color': tag.color() } : ''}
```
davwheat, line 20: Does this not change the functionality?

SychO9: it doesn't no, because we still only apply the color on CSS side when the element is active, see:
https://github.com/flarum/tags/pull/139/files/2ef572e7869b4f4d3b75a9a38156011967c627d7#diff-4a57d0becccd9b2141a8837b57fe71121d23b64f6f87884f3edabe4c56844071R33

js/src/forum/addTagFilter.js
SychO9: Done!

davwheat (approved):

PR #140, opened 2021-08-19
Fix discussions hidden from all users including admins

SychO9:
**Fixes flarum/core#3034**

**Reviewers should focus on**
Does this make sense ?


src/Access/GlobalPolicy.php
```diff
@@ -50,9 +50,11 @@ public function can(User $actor, string $ability)
             }
 
             if (! isset($enoughSecondary[$actor->id][$ability])) {
-                $enoughSecondary[$actor->id][$ability] = Tag::whereHasPermission($actor, $ability)
+                $secondaryTagCount = Tag::where('position', null)->count();
+
+                $enoughSecondary[$actor->id][$ability] = $ability === 'viewForum' && ($secondaryTagCount === 0 || Tag::whereHasPermission($actor, $ability)
```
askvortsov1, line 55: Doesn't this mean that for any ability other than viewForum the permission will be denied?

SychO9: yup.. 😅

askvortsov1:
I think the issue represents a more general case, where `$tagsWhereHasPermission < $tagsWhereNeedsPermission`, but only because `$tagsThatExist < $tagsWhereNeedsPermission`. Could we account for this by comparing to `>= min($tagCount, $tagsWhereNeedsPermission)`? Also, could we cache the counts statically?

SychO9:
That makes sense, and means that it also applies to the primary tags so we'd want to make the same change for them as well.

SychO9:
> Also, could we cache the counts statically?

That only executes once because `$enoughPrimary[$actor->id][$ability]` is cached statically, so I don't think there would be any value in doing so

src/Access/GlobalPolicy.php
```diff
@@ -44,15 +44,23 @@ public function can(User $actor, string $ability)
...
             if (! isset($enoughSecondary[$actor->id][$ability])) {
-                $enoughSecondary[$actor->id][$ability] = Tag::whereHasPermission($actor, $ability)
+                $secondaryTagsCount = Tag::query()->where(['position' => null, 'parent_id' => null])->count();
+                $secondaryTagsWhereNeedsPermission = $this->settings->get('flarum-tags.min_secondary_tags');
+                $secondaryTagsWhereHasPermission = Tag::whereHasPermission($actor, $ability)
                     ->where('tags.position', '=', null)
-                    ->count() >= $this->settings->get('flarum-tags.min_secondary_tags');
+                    ->count();
+
+                $enoughSecondary[$actor->id][$ability] = $secondaryTagsWhereHasPermission >= min($secondaryTagsCount, $secondaryTagsWhereNeedsPermission);
```
SychO9, line 63: hmm, I think we'd only want this specific logic to apply on the `viewForum` ability? if there are less tags than required, you should still not be able to `startDiscussion`. It's a bit hard to fully comprehend the tag abilities logic.

askvortsov1: Agreed, this should be for viewForum only, otherwise it should behave normally. If we keep it like this, and someone sets requirements to 0 for primary and secondary tags, all users will have all perms, which isn't right.

askvortsov1 (approved):

luceos:
I am going to require this logic is tested when `is_private` on a discussion is set to `1`, this could potentially be a test case. Please confirm that the private discussion is not showing up with this PR.

SychO9:
Added private discussions for each scenario, they still don't show up in the response.

tankerkiller125 (approved):

askvortsov1 (approved):

PR #141, opened 2021-08-20
Pass filter params

imorland:
Currently, any `filter` params that are passed to `Content\Tags` are ignored. As discussed with @askvortsov1 here is a solution...

Similar PR will be opened for core shortly :)

### DIFF
--- a/src/Content/Tag.php
+++ b/src/Content/Tag.php
@@ -63,6 +63,7 @@ class Tag
         $sort = Arr::pull($queryParams, 'sort');
         $q = Arr::pull($queryParams, 'q', '');
         $page = Arr::pull($queryParams, 'page', 1);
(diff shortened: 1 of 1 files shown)


src/Content/Tag.php
```diff
@@ -77,6 +78,8 @@ public function __invoke(Document $document, Request $request)
             'page' => ['offset' => ($page - 1) * 20, 'limit' => 20]
         ];
 
+        $params['filter'] = array_merge($params['filter'], $filters);
```
askvortsov1, line 81: $filters should override $params['filter'] on collision, so these should be swapped

askvortsov1 (changes requested):

src/Content/Tag.php
imorland: 🙈 whoops, apologies. Swapped

PR #145, opened 2021-09-14
Fix unauthorized view of restricted tags

SychO9:
**Fixes flarum/core#3050**

The tags in question could be seen because they were loaded as included relations, which didn't go through any visibility checks and therefore ended up in the response.

**Confirmed**

### DIFF
--- a/src/TagRepository.php
+++ b/src/TagRepository.php
@@ -14,6 +14,8 @@ use Illuminate\Database\Eloquent\Builder;
 
 class TagRepository
 {
+    private const TAG_RELATIONS = ['children', 'parent'];
+
     /**
      * Get a new query builder for the tags table.
      *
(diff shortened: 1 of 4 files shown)


src/TagRepository.php
```diff
@@ -24,6 +26,28 @@ public function query()
...
+    public function with($relations, User $actor): Builder
+    {
+        $relationsArray = is_string($relations) ? explode(',', $relations) : $relations;
+
+        foreach (self::TAG_RELATIONS as $relation) {
+            if (in_array($relation, $relationsArray, true)) {
+                $relationsArray = array_diff($relationsArray, [$relation]);
+
+                $relationsArray[$relation] = function ($query) use ($actor) {
+                    $query->whereVisibleTo($actor);
```
askvortsov1, line 43: Shouldn't there be a return here?

askvortsov1 (changes requested):

askvortsov1 (approved):
Tentatively approving as tests pass, brought up one potential concern above.

src/TagRepository.php
SychO9: nope, we just need to modify the query object.

tankerkiller125 (approved):

PR #146, opened 2021-09-29
fix: `$relationsArray` can end up containing closures

SychO9:
Not sure I didn't simply unset the first time around..

### DIFF
--- a/src/TagRepository.php
+++ b/src/TagRepository.php
@@ -33,15 +33,16 @@ class TagRepository
      */
     public function with($relations, User $actor): Builder
     {
-        $relationsArray = is_string($relations) ? explode(',', $relations) : $relations;
+        $relations = is_string($relations) ? explode(',', $relations) : $relations;
+        $relationsArray = [];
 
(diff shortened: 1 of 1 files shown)


src/TagRepository.php
```diff
@@ -35,9 +35,9 @@ public function with($relations, User $actor): Builder
     {
         $relationsArray = is_string($relations) ? explode(',', $relations) : $relations;
 
-        foreach (self::TAG_RELATIONS as $relation) {
+        foreach (self::TAG_RELATIONS as $k => $relation) {
```
askvortsov1, line 38: Maybe we should build a new array instead, and use that instead of reusing the one we were given? That feel safer in preventing issues like this.

askvortsov1 (changes requested):

src/TagRepository.php
SychO9: Sure

askvortsov1 (approved):

davwheat (approved):

PR #147, opened 2021-10-11
Lazy load tags in DiscussionTaggedPost if necessary

askvortsov1:
Closes flarum/core#3043

Depends on flarum/core#3100

Before: https://i.imgur.com/TaYra03.png

After: https://i.imgur.com/VB0GAVq.png

**Questions for Reviewers**

The `ShowTagController` change introduces ambiguity for fetching by ID vs slug, but I'm not sure that's avoidable, as the alternative is a `bySlug` query parameter, which would be a breaking change.


davwheat (changes requested):
Is there any reason we couldn’t add the bySlug query param anyway, but allow either without the param? That way, we can allow people to use it now to make it easier when we presumably deprecate it in 2.0.

datitisev:
What happens if the tag really is deleted? Will every DiscussionTaggedPost make an independent request on every `oninit`?

PR #150, opened 2021-11-05
Basic validation for tags slugs

askvortsov1:
**Fixes https://github.com/flarum/core/issues/3152**

Validate tags slugs to ensure they don't contain slashes or spaces.

I kept the requirements minimal for now, we can revisit additional constraints later if needed.

Unfortunately, even though the backend validation works, a `422` message won't actually be shown. This is because of https://github.com/flarum/core/issues/3154.

**Necessity**

**Confirmed**

- [x] Core developer confirmed locally this works as intended.

### DIFF
--- a/src/TagValidator.php
+++ b/src/TagValidator.php
@@ -18,7 +18,7 @@ class TagValidator extends AbstractValidator
      */
     protected $rules = [
         'name' => ['required'],
-        'slug' => ['required', 'unique:tags', 'regex:/^[^\/ ]*$/i'],
+        'slug' => ['required', 'unique:tags', 'regex:/^[^\/\\ ]*$/i'],
         'is_hidden' => ['bool'],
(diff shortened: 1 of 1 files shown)


src/TagValidator.php
```diff
@@ -18,7 +18,7 @@ class TagValidator extends AbstractValidator
      */
     protected $rules = [
         'name' => ['required'],
-        'slug' => ['required', 'unique:tags'],
+        'slug' => ['required', 'unique:tags', 'regex:/^[^\/ ]*$/i'],
```
SychO9, line 21: Backslashes create the same issue

```suggestion
        'slug' => ['required', 'unique:tags', 'regex:/^[^\/\\ ]*$/i'],
```

davwheat: Good catch

SychO9 (approved):

PR #154, opened 2022-01-05
Some fixes to tag metadata calculations

askvortsov1:
See commit descriptions.

Fixes https://github.com/flarum/core/issues/3252
Fixes https://github.com/flarum/core/issues/3262

### DIFF
--- a/src/Listener/UpdateTagMetadata.php
+++ b/src/Listener/UpdateTagMetadata.php
@@ -102,7 +102,7 @@ class UpdateTagMetadata
     {
         $discussion = $event->post->discussion;
         $delta = ! $discussion->exists && $discussion->hidden_at === null ? -1 : 0;
-        $this->updateTags($event->post->discussion, $delta);
+        $this->updateTags($discussion, $delta);
     }
 
     /**
--- a/src/Listener/UpdateTagMetadata.php
+++ b/src/Listener/UpdateTagMetadata.php
@@ -100,7 +100,9 @@ class UpdateTagMetadata
      */
     public function whenPostIsDeleted(PostDeleted $event)
     {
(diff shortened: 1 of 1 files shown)


src/Listener/UpdateTagMetadata.php
```diff
@@ -100,7 +100,9 @@ public function whenPostIsPosted(Posted $event)
      */
     public function whenPostIsDeleted(PostDeleted $event)
     {
-        $this->updateTags($event->post->discussion);
+        $discussion = $event->post->discussion;
+        $delta = ! $discussion->exists && $discussion->hidden_at === null ? -1 : 0;
+        $this->updateTags($event->post->discussion, $delta);
```
SychO9, line 105:
```suggestion
        $this->updateTags($discussion, $delta);
```

src/Listener/UpdateTagMetadata.php
```diff
@@ -100,7 +100,9 @@ public function whenPostIsPosted(Posted $event)
      */
     public function whenPostIsDeleted(PostDeleted $event)
     {
-        $this->updateTags($event->post->discussion);
+        $discussion = $event->post->discussion;
+        $delta = ! $discussion->exists && $discussion->hidden_at === null ? -1 : 0;
```
SychO9, line 104: so, if the discussion has been permanently deleted and *not* soft deleted? wouldn't it be enough to check that the discussion is permanently deleted?

askvortsov1: If the discussion is soft deleted, then the counter has already been decreased (since discussion hide causes the delta to decrease), and so we shouldn't decrease it again.

SychO9 (approved):

PR #156, opened 2022-01-24
Export components so they can be extended.

grimmdude:

I have a use case to extend these components, but since they're not exported they're not easily extended.

Ensure the extension still builds as expected.


SychO9:
Any specific components you're unable to access? because most of them are exported already through https://github.com/flarum/tags/blob/master/js/src/forum/compat.js

you would import for example like so: `import TagHero from 'flarum/tags/TagHero'`

grimmdude:
Thanks @SychO9, currently I'm looking to specifically extend `TagsPage` (forum).  I was following the documentation on importing from extensions and this seemed to be the way other extensions were handling it.  

https://docs.flarum.org/extend/extending-extensions#importing-from-extensions

However, when I try importing using your suggestion I'm not having any luck.  To be clear, I'm importing from _another_ extension, not directly within the app itself.  Let me know if I'm missing something though 🙏

datitisev:
@grimmdude I believe that system should be used for non-core extensions. I might be wrong though. What Sami mentioned should work for you though.

grimmdude:
@datitisev got it thanks, that would make sense that the other method would be for non-core extensions.  I was able to get that import to work with:

```js
import TagsPage from 'flarum/tags/forum/components/TagsPage';
```

Thanks to you both!

PR #161, opened 2022-02-20
Fix tag page mismatches if accents/capitalizations differ in slug

datitisev:
**Fixes flarum/core#3286**

- Use `localeCompare` to match tag slugs regardless of capitalization & accents
- Add `this.currentTagLoading` to prevent multiple requests if a HTTP req needs to be made to find the tag 
  - See below in focus areas

I don't know if this is the best way to do it. I think its preferred to `toLowerCase` though, as it deals with accents.

While investigating these changes, I found that the backend doesn't find models with some accents (e.g. äa).

I don't love the `currentTagLoading` thing but I think that whole system needs to be fixed separately. We also probably need to address the infinite requests if the user goes to a non-existent tag with JS.

**Screenshot**
Slug is `äa` (which gives 404 on refresh, but just to show the frontend changes)
image

**Necessity**

- [X] Has the problem that is being solved here been clearly explained?
- [ ] If applicable, have various options for solving this problem been considered?
- [X] For core PRs, does this need to be in core, or could it be in an extension?
- [ ] Are we willing to maintain this for years / potentially forever?

**Confirmed**

- [X] Frontend changes: tested on a local Flarum installation.
- ~~Backend changes: tests are green (run `composer test`).~~
- [ ] Tests have been added, or are not appropriate here.

**Required changes:**

- [ ] Related documentation PR: (Remove if irrelevant)
- [ ] Related core extension PRs: (Remove if irrelevant)

### DIFF
--- a/js/src/forum/addTagFilter.js
+++ b/js/src/forum/addTagFilter.js
@@ -6,6 +6,8 @@ import classList from 'flarum/utils/classList';
 
 import TagHero from './components/TagHero';
 
+const findTag = slug => app.store.all('tags').find(tag => tag.slug().localeCompare(slug, undefined, { sensitivity: 'base' }) === 0);
+
 export default function() {
   IndexPage.prototype.currentTag = function() {
     if (this.currentActiveTag) {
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
LGTM

davwheat (approved):

PR #164, opened 2023-04-09
Update tag filtering to include partial matches

rafaucau:
This pull request updates the tag filtering functionality to include partial matches when searching for tags. The previous implementation used `substring` method to match the tags with the filter string, which only matched tags starting from the beginning of the tag name.

The new implementation uses the `includes` method to match the filter string with any part of the tag name. This means that users can now search for tags based on partial matches, making it easier to find relevant tags.

**Screenshot**

Tag list:
image

| Before | After |
| ------ | ----- |
| Before | After |

**Necessity**

- [x] For core PRs, does this need to be in core, or could it be in an extension?

**Confirmed**

### DIFF
--- a/js/src/common/components/TagSelectionModal.tsx
+++ b/js/src/common/components/TagSelectionModal.tsx
@@ -260,7 +260,7 @@ export default class TagSelectionModal<
     // If the user has entered text in the filter input, then filter by tags
     // whose name matches what they've entered.
     if (filter) {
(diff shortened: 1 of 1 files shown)


davwheat (approved):

SychO9:
Thanks for the PR!

As this is a read-only repository I will be closing this, the actual development takes place at https://github.com/flarum/framework, please feel free to re-open the PR there 🙏🏼.

rafaucau:
Re-opened https://github.com/flarum/framework/pull/3786

PR #165, opened 2025-05-14
Fixed issue https://github.com/flarum/framework/issues/3692

edgeinfinity1:
**Fixes #0000**

Fixing an issue causing users unable to rename their posts, which has been confirmed at https://github.com/flarum/framework/issues/3692

This is a rather brutal fix, since I'm not too into flarum. Please consider rewriting this in an elegant way.
At least I have confirmed it working on my own deployment at https://forum.monika.love

**Screenshot**

**Necessity**

- [x] For core PRs, does this need to be in core, or could it be in an extension?

**Confirmed**

- [ ] Frontend changes: tested on a local Flarum installation.

### DIFF
--- a/src/Access/DiscussionPolicy.php
+++ b/src/Access/DiscussionPolicy.php
@@ -35,6 +35,21 @@ class DiscussionPolicy extends AbstractPolicy
             foreach ($tags as $tag) {
                 if ($tag->is_restricted) {
                     if (! $actor->hasPermission('tag'.$tag->id.'.discussion.'.$ability)) {
+
+                        if ($ability == 'rename') {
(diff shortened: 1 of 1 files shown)


luceos:
Thank you for your contribution @edgeinfinity1.

For Flarum we work with the subsplit method - similar to Laravel. This means that the current repository is read only, all code changes and pull request need to target the main repository: https://github.com/flarum/framework. Your contribution, as such needs to target this directory: https://github.com/flarum/framework/tree/2.x/extensions/tags

We would appreciate if you could retarget your contribution there.

Thanks again, and have a great weekend.

